PR #1, opened 2022-05-22
Missile testing

ky28059:
Perhaps a bit unconventional, but there are swerve module and project changes on this branch which can be merged after missile testing concludes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,6 +81,10 @@ jar {
     duplicatesStrategy = DuplicatesStrategy.INCLUDE
 }
 
+test {
+    useJUnit()
+}
+
 // Configure jar and deploy tasks
 deployArtifact.jarTask = jar
 wpi.java.configureExecutableTasks(jar)
--- a/src/main/java/frc/robot/Main.java
+++ b/src/main/java/frc/robot/Main.java
@@ -12,14 +12,14 @@ import edu.wpi.first.wpilibj.RobotBase;
  * call.
  */
 public final class Main {
-  private Main() {}
+    private Main() {}
 
-  /**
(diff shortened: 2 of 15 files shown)


ky28059:
image

Leaving this here so I don't forget over the summer:

### Steps to tune swerve
#### On the first missile module (using `MissileShellSwerveSubsystem`, percent output instead of m/s velocity):
1. Check module radian conversion factor
2. Tune module steer PID
3. Check to make sure everything is in working order (continuous rotation in both directions, angle modulus, swerve state optimization)

#### On the robot (using `SwerveSubsystem`, percent output, robot on wood initially):
1. Put CAN IDs in `Constants` and pass module positions (relative to the robot center, in meters) to `SwerveDriveKinematics`
2. Tune module offset by setting offset in code to 0, rotating it by hand (after code redeploy) to the desired zero, and checking shuffleboard for the module angle in radians -- the offset should be the inverse of the angle. If the angle is correct but the wheel spins backwards, add `Math.PI` to the offset
3. Test rotation, tune `MAX_OMEGA` (?)
4. Take it for a test drive!

#### Final steps? (using `SwerveSubsystem`, percent output initially)
1. Tune `DRIVE_ROTATIONS_TO_METERS` by pushing robot a known distance in meters
2. Run robot on wood and check shuffleboard for max power m/s speed for `MAX_VEL`
3. Tune `MAX_OMEGA` again?

ky28059:
Merged because the main goal - running the 2020 robot - was achieved (at least for a few days), and everything was set up for running 2023 on the missile when a module is available. Further troubleshooting of the problems later encountered with 2020 will probably go on a different branch, if they are deemed necessary at all.

PR #2, opened 2022-07-27
Swerve module subclasses

ky28059:
This is the 4 classes configuration described in https://discord.com/channels/964689311540838471/964689312115482656/1001301901587001364, which should allow each module to be labelled with only one number (the pin offset) with the constructor handling the additional 90 degree offset.

### DIFF
--- a/src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
+++ b/src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
@@ -53,7 +53,6 @@ public class SwerveModule {
      */
     public SwerveModule(int drivePort, int steerPort, double offsetRads) {
         driveMotor = new GRTTalonFX(drivePort);
-        driveMotor.configFactoryDefault();
         driveMotor.setNeutralMode(NeutralMode.Brake);
 
         driveMotor.configSelectedFeedbackSensor(FeedbackDevice.IntegratedSensor);
--- a/src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
+++ b/src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
@@ -147,4 +147,56 @@ public class SwerveModule {
     private Rotation2d getAngle() {
         return new Rotation2d(steerEncoder.getPosition() + offsetRads);
     }
+
+    /**
+     * Utility class to construct a top left swerve module from a module's pin offset.
+     * The offset to align the pin with the front of the robot is automatically applied.
+     */
+    public static class TopLeft extends SwerveModule {
+        public TopLeft(int drivePort, int steerPort) {
+            super(drivePort, steerPort, 0.0);
+        }
+        public TopLeft(int drivePort, int steerPort, double offsetRads) {
+            super(drivePort, steerPort, offsetRads);
+        }
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
```diff
@@ -147,4 +147,56 @@ public static Pair<Double, Double> optimizeModuleState(SwerveModuleState target,
...
+        public TopLeft(int drivePort, int steerPort, double offsetRads) {
+            super(drivePort, steerPort, offsetRads);
+        }
+    }
+
+    /**
+     * Utility class to construct a top right swerve module from a module's pin offset.
+     * The offset to align the pin with the front of the robot is automatically applied.
+     */
+    public static class TopRight extends SwerveModule {
```
e3l, line 168: would it be cleaner to use an enum in SwerveModule's constructor instead of having four subclasses?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -33,10 +33,10 @@ public class SwerveSubsystem extends SubsystemBase {
...
     public SwerveSubsystem() {
         // Initialize swerve modules
-        topLeftModule = new SwerveModule(tlDrive, tlSteer);
-        topRightModule = new SwerveModule(trDrive, trSteer);
-        bottomLeftModule = new SwerveModule(blDrive, blSteer);
-        bottomRightModule = new SwerveModule(brDrive, brSteer);
+        topLeftModule = new SwerveModule.TopLeft(tlDrive, tlSteer);
+        topRightModule = new SwerveModule.TopRight(trDrive, trSteer);
+        bottomLeftModule = new SwerveModule.BottomLeft(blDrive, blSteer);
+        bottomRightModule = new SwerveModule.BottomRight(brDrive, brSteer);
```
e3l, line 39: add variables in Constants.java?

e3l (commented):
some suggestions

src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
ky28059: Perhaps, but this way we can typecheck the alignment of the modules in `SwerveSubsystem` so that it generates a compile time error of someone accidentally constructs a top right module for the top left module variable.

ky28059: Maybe if that's not necessary an enum would too.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
ky28059: Like the offsets? The CAN IDs are already in `Constants`.

e3l: Yeah, the offsets, sorry

src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
e3l: Hmm, I guess that's nice. Either way it doesn't matter too much, it's compile time errors vs a couple less class declarations, so it's honestly up to you

PR #3, opened 2022-11-03
Swerve subsystem locking

ky28059:
Not sure if this is the most ideal implementation, so I'm putting it on a PR for now.

### DIFF
--- a/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
@@ -7,11 +7,11 @@ import edu.wpi.first.math.Nat;
 import edu.wpi.first.math.estimator.SwerveDrivePoseEstimator;
 import edu.wpi.first.math.geometry.Pose2d;
 import edu.wpi.first.math.geometry.Rotation2d;
-import edu.wpi.first.math.geometry.Translation2d;
 import edu.wpi.first.math.kinematics.ChassisSpeeds;
 import edu.wpi.first.math.kinematics.SwerveDriveKinematics;
 import edu.wpi.first.math.kinematics.SwerveModuleState;
 import edu.wpi.first.wpilibj.SPI;
+import edu.wpi.first.wpilibj.Timer;
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
 
 import static frc.robot.Constants.SwerveConstants.*;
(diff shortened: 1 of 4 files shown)


src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -121,6 +125,20 @@ public void periodic() {
         );
     }
 
+    /**
+     * Toggles whether the subsystem is locked (whether control input has no effect and the wheels 
+     * should lock themselves to prevent shoving).
+     */
+    public void toggleLocked() {
```
e3l, line 132: I think our software design is intended to avoid actually interfacing with motors as a result of button presses, but rather changing states and allowing each module to interface with its motors as part of its periodic loop. 

so I think this motor setting logic should go in periodic

e3l: another thought is that the correct logic might not be to immediately turn the modules. we might want to put in logic to lock only after a certain delay, or after the robot is going at less than a certain velocity, even if we dont use it?

i know in FTC, the one team that built a swerve robot could immediately lock their wheels and when they go full speed and let go of the joystick they nearly flip over

e3l (commented):
seems like it would work, but some nitpicky comments about design / possible features we'd want

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -31,6 +31,19 @@ public class SwerveSubsystem extends SubsystemBase {
     public static final double MAX_ACCEL = 3; // Max robot tangential acceleration, in m/s^2
     public static final double MAX_OMEGA = Math.toRadians(30); // Max robot angular velocity, in rads/s
 
+    private final Timer lockTimer;
+    private static final double LOCK_TIMEOUT_SECONDS = 1.0; // The elapsed idle time to wait before locking
+    private static final boolean LOCK_TIMEOUT_ENABLE = true;
```
e3l, line 36: possible misnomer? seems to be whether to lock at all

e3l (commented):
lgtm with 1 comment

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
ky28059: Perhaps; might be better as `LOCKING_ENABLE` then?

ky28059:
Going to merge because of missile testing; if there's an issue with locking we can fix it on `main`.

PR #5, opened 2022-11-06
Use relative encoder for swerve steering

ky28059:
Hopefully addresses https://grt2023.slack.com/archives/C03U2MQE4C8/p1667623458767849.

### DIFF
--- a/src/main/java/frc/robot/subsystems/swerve/MissileShellSwerveSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/swerve/MissileShellSwerveSubsystem.java
@@ -21,7 +21,7 @@ public class MissileShellSwerveSubsystem extends SubsystemBase {
     };
 
     public MissileShellSwerveSubsystem() {
-        module = new SwerveModule(2, 1, -1.1350287199020386);
+        module = new SwerveModule(2, 1, -1.1350287199020386 + Math.PI);
 
         // One module at the center of the robot
         kinematics = new SwerveDriveKinematics(
--- a/src/main/java/frc/robot/subsystems/swerve/MissileShellSwerveSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/swerve/MissileShellSwerveSubsystem.java
@@ -33,9 +33,11 @@ public class MissileShellSwerveSubsystem extends SubsystemBase {
     private final SwerveDriveKinematics kinematics;
 
     public static final double MAX_VEL = 1.0; // Max robot tangential velocity, in percent output
-    private static final double offsetRads = 0;
+    private static final double offsetRads = -1.1350287199020386;
 
-    private static final double STEER_ROTATIONS_TO_RADIANS = (1.0 / 39.0) * (34.0 / 63.0) * 2 * Math.PI; // 39:1 gear ratio, 63:34 pulley ratio, 1 rotation = 2pi
+    // private static final double STEER_ROTATIONS_TO_RADIANS = (1.0 / 39.0) * (34.0 / 63.0) * 2 * Math.PI; // 39:1 gear ratio, 63:34 pulley ratio, 1 rotation = 2pi
+    private static final double STEER_ROTATIONS_TO_RADIANS = (1 / 94.65) * 2 * Math.PI; // 39:1 gear ratio, 63:34 pulley ratio, 1 rotation = 2pi
+
     private static final double STEER_VOLTS_TO_RADIANS = 2 * Math.PI / 3.3; // MA3 analog output: 3.3V -> 2pi
 
     private static final double steerP = 0.4;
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/subsystems/swerve/MissileShellSwerveSubsystem.java
```diff
@@ -51,11 +56,15 @@ public MissileShellSwerveSubsystem() {
...
+        steerAbsoluteEncoder = steerMotor.getAnalog(SparkMaxAnalogSensor.Mode.kAbsolute);
+        steerAbsoluteEncoder.setPositionConversionFactor(STEER_VOLTS_TO_RADIANS);
+
+        steerEncoder = steerMotor.getEncoder();
+        steerEncoder.setPositionConversionFactor(STEER_ROTATIONS_TO_RADIANS);
+        steerEncoder.setPosition(steerAbsoluteEncoder.getPosition()); // Set initial position to absolute value
 
         steerPidController = steerMotor.getPIDController();
-        steerPidController.setFeedbackDevice(steerEncoder);
+        //steerPidController.setFeedbackDevice(steerAbsoluteEncoder);
```
e3l, line 67: does this not need to be called on the built-in encoder

src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
```diff
@@ -25,12 +26,15 @@ public class SwerveModule {
...
     private final CANSparkMax steerMotor;
-    private final SparkMaxAnalogSensor steerEncoder;
+    private final RelativeEncoder steerEncoder;
+    private final SparkMaxAnalogSensor steerAbsoluteEncoder;
     private final SparkMaxPIDController steerPidController;
 
     private final double offsetRads;
 
     private static final double DRIVE_TICKS_TO_METERS = 1.0;
+    private static final double STEER_ROTATIONS_TO_RADIANS = 1.0 / 39.0 * 2 * Math.PI; // 39:1 gear ratio, 1 rotation = 2pi
```
e3l, line 36: jacob sent an update on this constant:

> the gearbox has a ratio of 39:1
> but the pulleys have a ratio of 63:34

e3l (changes requested):

src/main/java/frc/robot/subsystems/swerve/MissileShellSwerveSubsystem.java
ky28059: It defaults to the built-in relative encoder. We actually didn't use this method at all last year; I only discovered it while researching analog feedback.

src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
ky28059: So it'd be `1.0 / 39.0 * 34.0 / 63.0` for neo rotations to module rotations?

e3l: i believe so, yes. with parenthesis

e3l (commented):
looks good. why can't we consolidate the two classes (MissileShellSwerveSub and SwerveSub), by the way?

ky28059:
`SwerveSubsystem` is 4 modules on a base, `MissileShellSwerveSubsystem` is 1 module on the missile to tune constants and offsets. The latter probably won't be needed later in the season.

e3l:
Ah, could we have MSSS construct a SwerveSub and then set states? I'm just worried abt copying logic back and forth

ky28059:
If you mean construct a `SwerveModule`, the old logic used to do that, but because we're using double NEOs for our test modules and `SwerveModule` still assumes NEO and Falcon the logic isn't entirely the same. I can update `SwerveModule` to use double NEOs in the meantime maybe?

e3l:
Oh, yes I see that now, sorry. Perhaps we can just stick to double NEOs and transition for good once we have Falcons, yeah.. because switching it back doesn't involve too much logic? 

Or we could put a flag in SwerveSubsystem to switch between the two types of drive motors, if we see ourselves switching between the two more...

ky28059:
I think being able to switch between the two is a bit complex, since the flag can't exist at the `SwerveModule` level because falcons and NEOs need different classes to be constructed and stored in fields in order to control them (NEO needs the spark, encoder, and pidcontroller, while falcon just needs the talon) and dynamically populating these seems like a bad idea in java.

If we wanted switching, the best way to do that would probably be to have an abstract class `BaseSwerveModule` that implements steering and declares the set / get state API, and then two subclasses that construct either a TalonFX or CANSparkMax for the drive.

If we can avoid it though, it would probably be best if all our modules were homogenous and we didn't have to swap between the two motors constantly. I'll work on changing `SwerveModule` to use double NEOs for now.

e3l (approved):
lgtm! we'll have to add back shuffleboard stuff to tune PIDs later, right

ky28059:
I'm holding off on shuffleboard for now because there might be a better way to do debugging that isn't how it's currently set up. I'll merge this and probably move shuffleboard to a separate PR.

PR #6, opened 2022-11-16
Swerve module debug flag

ky28059:
Not sure how best to set this up, since swerve modules aren't subsystems and therefore don't have periodic loops to set shuffleboard entries. Currently the entry values are set in `setDesiredState()`, but we could also consider having a debug method for the subsystem to print, not using shuffleboard in favor of `System.out.println()`, etc.


e3l (approved):
lgtm

sorry, i don't get what you were saying abt the debug method for printing, but can't that go in an if statement too?

e3l:
Is this still applicable?

ky28059:
No. I'll close this PR and if we need shuffleboard capability down the line it'll need to be rewritten for 2023 anyways.

PR #8, opened 2022-11-16
Swerve idle logic

### LINKED ISSUE #7: Swerve TODOs
Some tasks before I forget that we can test on the trike before being fully happy with our setup.

- [ ] Not returning to 0 degrees when the joystick is released. This probably needs a special case in the swerve subsystem's `setSwerveDrivePowers()` method that checks if all powers supplied are zero, and if so to set the module states to their current angle, but 0 velocity. Not sure if querying all 4 states in the method and just mutating those classes is super clean though.
- [ ] Timer-based locking. This logic is already written and just needs to be tested on the 2020 robot (whether the actual positions work to lock the robot has already been tested).

ky28059:
I'm not super in love with how the "don't return to 0" logic is handled, but I'm not seeing too many alternatives or methods of simplification at the moment. Definitely will need some thought / testing.

Closes #7.

### DIFF
--- a/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
@@ -83,16 +83,10 @@ public class SwerveSubsystem extends SubsystemBase {
     public void setSwerveDrivePowers(double xPower, double yPower, double angularPower) {
         // If drivers are sending no input, stop all modules but hold their current angle.
         if (xPower == 0.0 && yPower == 0.0 && angularPower == 0.0) {
-            SwerveModuleState tlState = topLeftModule.getState();
-            SwerveModuleState trState = topRightModule.getState();
-            SwerveModuleState blState = bottomLeftModule.getState();
-            SwerveModuleState brState = bottomRightModule.getState();
-
-            this.states[0] = new SwerveModuleState(0.0, tlState.angle);
-            this.states[1] = new SwerveModuleState(0.0, trState.angle);
-            this.states[2] = new SwerveModuleState(0.0, blState.angle);
-            this.states[3] = new SwerveModuleState(0.0, brState.angle);
-
+            this.states[0] = new SwerveModuleState(0.0, this.states[0].angle);
+            this.states[1] = new SwerveModuleState(0.0, this.states[1].angle);
+            this.states[2] = new SwerveModuleState(0.0, this.states[2].angle);
+            this.states[3] = new SwerveModuleState(0.0, this.states[3].angle);
             return;
         }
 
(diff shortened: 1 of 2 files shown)


ky28059:
I added `isIdle` as a state because I thought that setting the state repeatedly while idle was bad, but realized that since we set the state repeatedly during normal controller operation that doing so while idle is probably fine (it doesn't interfere with locking since locking ignores the set states while active). The only problem I might foresee is that repeatedly querying the encoder might be a performance concern? but that will just have to be tested.

At the very least, the `isIdle` that is set in `setSwerveDrivePowers()` *cannot* be reused in the periodic loop to check to see if we should start the lock timer, since auton sets the system states through a different method and I think our current setup would just remain locked during auton (or at least not lock during auton if we're just standing still).

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -81,6 +81,21 @@ public SwerveSubsystem() {
...
      */
     public void setSwerveDrivePowers(double xPower, double yPower, double angularPower) {
+        // If drivers are sending no input, stop all modules but hold their current angle.
+        if (xPower == 0.0 && yPower == 0.0 && angularPower == 0.0) {
+            SwerveModuleState tlState = topLeftModule.getState();
+            SwerveModuleState trState = topRightModule.getState();
+            SwerveModuleState blState = bottomLeftModule.getState();
+            SwerveModuleState brState = bottomRightModule.getState();
+
+            this.states[0] = new SwerveModuleState(0.0, tlState.angle);
```
ky28059, line 91: I also just realized that we can use the previous setpoint's angle (`states[0].angle`) instead of the queried angle to prevent one call to the encoder. Is this better?

e3l (commented):
Is there a particular reason this logic is in the state setter rather than in periodic? 

I think I liked when it used the encoder position, because that seems to be the most 'lazy' method possible for this 'lazy' logic. Aren't we querying the encoders in periodic already anyways?

Additionally, will there ever be a case where we want to to set the modules to prepare for a movement, without actually performing that movement? For example, if we're stopping in auto and want to prepare our next movement while the robot does other things?

ky28059:
1. The logic is in the state setter because we can't actually know in `periodic()` whether the driver input is idle. We can maybe guess by checking if all the setpoint velocities are `0.0` and the angles are 0 degrees, but then we don't know what the previous setpoints are to set the modules to. I guess it could work if in `periodic()` we check the speeds for idle and ignore the `states` variable using the encoder readings for the angle setpoints? I also just think it makes sense in the state setter because drivers releasing the joystick causing the setpoint to be 0 degrees isn't really the drivers commanding the modules to go to 0 degrees, so I think it's fair if we set `states` to just what the drivers requested which is 0 velocity.
2. We are, but an additional query (in a non-periodic method) is additional overhead right? I guess the solution described above would fix this.
3. Possibly, but `setSwerveDrivePowers()` is not invoked during auto (default command) so the idle logic doesn't apply there. I guess maybe if the lock timeout kicks in at an unfortunate time it might be bad?

ky28059:
Oh, but if we have the idle logic in periodic, if we want to set the modules to a 0 velocity set angle state during auton it won't work.
```java
} else if (isIdle) {
    topLeftModule.setDesiredState(new SwerveModuleState(0.0, tlState.angle));
    topRightModule.setDesiredState(new SwerveModuleState(0.0, trState.angle));
    bottomLeftModule.setDesiredState(new SwerveModuleState(0.0, blState.angle));
    bottomRightModule.setDesiredState(new SwerveModuleState(0.0, brState.angle));
}
```

PR #12, opened 2022-11-19
Field orient using estimator value instead of gyro

ky28059:
Handles resetting the swerve field-oriented 0 degree angle by calling `resetPosition()`. We can't zero the gyro to reset the robot angle because doing so will mess up the pose estimator.

image

Definitely need to test this to make sure we're using the pose estimator correctly, though if we properly tune it this might even be more accurate than just using the gyro.


e3l:
I think we should save this merge for when we fully tune, test, and integrate the pose estimator, probably on our first swerve dt.

e3l:
It seems like we should use only gyro to zero the swerve field-oriented drive, and have a method to reset this gyro's angle (perhaps not by resetting the gyro, but by storing an offset double as a field variable).

e3l:
Closing in favor of using gyro for field-oriented navigation

PR #13, opened 2022-11-19
Field resetting for FLL demo

ky28059:
I put the reset on the drive B button for now, let me know if another button is preferable.


e3l:
I don't think we have time tonight to tune (and test) poseestimator, which also currently uses the powers set to the modules to estimate robot position (which i imagine will be quite innaccurate).

Can we ditch the pose estimator and just use the gyro for now?

e3l:
never got to test this on the 2020 robot, let's just integrate and test this on our 2023 swerve

PR #14, opened 2022-11-25
4th Shop Project

e3l:
PR for comments


src/main/java/frc/robot/Constants.java
```diff
@@ -42,4 +42,26 @@ public static final class SwerveConstants {
     public static final class ShuffleboardConstants {
         public static final double UPDATE_TIME = 0.5; // seconds between each update
     }
+
+    public static final class TankConstants {
+        public static final int motorBackLeft = 13;
```
e3l, line 47: java constants are ALL_CAPS

src/main/java/frc/robot/Robot.java
```diff
@@ -34,6 +34,7 @@ public void robotPeriodic() {
         // and running subsystem periodic() methods. This must be called from the robot's periodic
         // block in order for anything in the Command-based framework to work.
         CommandScheduler.getInstance().run();
+        robotContainer.periodic();
```
e3l, line 37: for practice, let's make sure that this only runs in teleop. We don't want states to be overwritten during auto

src/main/java/frc/robot/subsystems/AlignerSubsystem.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class AlignerSubsystem extends SubsystemBase {
+    private final WPI_TalonSRX motorSlapper = MotorUtil.createTalonSRX(slapId);
+    private final WPI_TalonSRX motorAngler = MotorUtil.createTalonSRX(angleId);
+    private final GRTShuffleboardTab shuffleboardTab = new GRTShuffleboardTab("Aligner");
+    private final GRTNetworkTableEntry positionEntry = shuffleboardTab.addEntry("Slapperposition", 0).at(0, 0); 
+
+    public double state = 0;
+    public double slapper;
+    public double angler;
```
e3l, line 22: i think we should adopt the convention of putting states at the very top, for readability

e3l (commented):
looks like both aligner and elevator need a little more code

src/main/java/frc/robot/subsystems/TankSubsystem.java
```diff
@@ -0,0 +1,52 @@
...
+        leftmotor2.setNeutralMode(NeutralMode.Brake);
+        rightmotor.setNeutralMode(NeutralMode.Brake);
+        rightmotor2.setNeutralMode(NeutralMode.Brake);
+
+
+        leftmotor2.follow(leftmotor);
+        rightmotor2.follow(rightmotor);
+
+        leftmotor.setInverted(true);
+        leftmotor2.setInverted(true);
```
ky28059, line 35: You can use `setInverted(InvertType)` as a cleaner way of making a follower have the same inversion as the main. See [last year's shop project code](https://github.com/grt192/2021ShopProj/blob/6thPerClawMech/src/main/java/frc/robot/subsystems/tank/TankSubsystem.java#L34) and the [CTRE docs](https://docs.ctre-phoenix.com/en/latest/ch13_MC.html#follower).

src/main/java/frc/robot/subsystems/AlignerSubsystem.java
```diff
@@ -0,0 +1,103 @@
...
+        else if (to_grab && angler_open) {
+            current_slaptarget = CLOSEDSLAP;
+            current_anglertarget = CLOSEDANGLER;
+            angler_open = false;
+        }
+        // check if grab trigger is pressed and angler is closed
+        else if (to_grab && !angler_open) {
+            current_slaptarget = OPENSLAP;
+            current_anglertarget = OPENANGLER;
+            angler_open = true;
```
e3l, line 72: wouldn't these two states just flip hella fast if the grab trigger is held down?

i also don't really get angler_open, how is it different from to_grab? I think it would make more sense to put toggling logic in RobotContainer so that all the states in this subsystem directly represent what we want the hardware to be doing, rather than like a button state (that we later process using toggle logic).

src/main/java/frc/robot/subsystems/AlignerSubsystem.java
```diff
@@ -0,0 +1,103 @@
...
+        // set motors to brake
+        motorSlapper.setNeutralMode(NeutralMode.Brake);
+        motorAngler.setNeutralMode(NeutralMode.Brake);
+
+        // setup encoders for motors
+        motorSlapper.configSelectedFeedbackSensor(FeedbackDevice.QuadEncoder);
+        motorSlapper.setSensorPhase(true);
+
+        motorAngler.configSelectedFeedbackSensor(FeedbackDevice.QuadEncoder);
+        motorAngler.setSensorPhase(true);
```
e3l, line 45: usually best practice to also reset the encoder positions to 0 (just in case, i guess)
relative encoders (which we're using) are only capable of measuring rotations (well, 'ticks') since boot, so we'll need to make sure that the aligners start in a known 0 position and all our measurements are relative to that position

e3l (commented):
comments for aligners

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    public Height height;
+
+    public ElevatorSubsystem() {
+        height = Height.GROUND;
+
+        motorElevator.configSelectedFeedbackSensor(FeedbackDevice.QuadEncoder);
+        motorElevator.setSensorPhase(true);
```
e3l, line 68: same comment as https://github.com/grt192/GRT2023/pull/14#discussion_r1039020378

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public void periodic() {
+        elevatorPositionEntry.setValue(motorElevator.getSelectedSensorPosition());
+        
+        if(motorElevator.getSelectedSensorPosition() > height.get() + WINCHTOLERANCE){
+            motorElevator.set(WINCHSPEED); // WARNING: IF THIS MOTOR IS INVERTED, THIS WILL GO TERRIBLY WRONG
+            // MAKE SURE THAT THIS GOES THE RIGHT WAY BEFORE TESTING THOROUGHLY
+        } else if(motorElevator.getSelectedSensorPosition() < height.get() - WINCHTOLERANCE){
+            motorElevator.set(WINCHSPEED * -1);
```
e3l, line 78: we might need to use PID, so i'd say read into it to make sure you at least can swap it if we find we need it

but also, this logic never sets the motor to stop moving. also be sure to use the proper idle behavior

e3l (commented):
comments for elevator

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private final GRTShuffleboardTab shuffleboardTab = new GRTShuffleboardTab("Elevator");
+    private final GRTNetworkTableEntry elevatorPositionEntry = shuffleboardTab.addEntry("ElevatorPosition", 0).at(0, 0);
+
+    public enum Height {
+        GROUND {
+            public Height previous() {
+                return Height.GROUND;
+            }
+            public double get(){
```
ky28059, line 27: I suppose this is in the past this year, but for future reference you can use an enum with a private constructor and a value field instead of having to override `.get()` for everything.
```java
public enum Height {
    GROUND(GROUND_HEIGHT), LOW(LOW_HEIGHT), HIGH(HIGH_HEIGHT);

    public double value;
    private Height(double height) {
        this.value = height;
    }
}
```
This is what we did last year for [intake positions](https://github.com/grt192/GRT2022/blob/develop/src/main/java/frc/robot/subsystems/IntakeSubsystem.java#L31-L43).

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+        
+        double distance = height.get() - motorElevator.getSelectedSensorPosition();
+
+        // System.out.println(motorElevator.getSelectedSensorPosition());
+        // motorElevator.set(speed);
+
+
+
+        
+        if(Math.abs(distance) > WINCHTOLERANCE){
```
ky28059, line 100: Should use PID for this in the future, especially because PID is going to be everywhere during build.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+        
+        motorElevator.configForwardSoftLimitThreshold(WINCHMAX, 0);
+        motorElevator.configForwardSoftLimitEnable(true, 0);
+
+        motorElevator.configReverseSoftLimitThreshold(0, 0);
+        motorElevator.configReverseSoftLimitEnable(true, 0);
+
+    }
+
+    public double speedfromdistance(double x){
```
ky28059, line 84: This function seems like a weird custom PID loop, and doesn't seem to be used?

src/main/java/frc/robot/subsystems/AlignerSubsystem.java
```diff
@@ -0,0 +1,127 @@
...
+            
+        // else return to open state
+        else {
+            current_slaptarget = OPENSLAP;
+            current_anglertarget = OPENANGLER;
+        }
+        System.out.println(current_slaptarget);
+
+        // get slapper to current target
+        if (Math.abs(slappos - current_slaptarget) <= motorrange) {
```
ky28059, line 105: Seems like PID would be more ideal for this as well.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -74,19 +64,69 @@ public RobotContainer() {
...
     /**
      * Use this to pass the autonomous command to the main {@link Robot} class.
+     * 
      * @return the command to run in autonomous
      */
     public Command getAutonomousCommand() {
         return autonChooser.getSelected();
     }
+
+    public void periodic() {
```
ky28059, line 78: Default commands :( they're easier to read! Plus less confusing since you technically want this method to be a `teleopPeriodic()`, not just a regular `periodic()` loop that also runs during auton. Plus using button callbacks handles debounce and other oddities; doing it manually just seems like theres a lot more room for problems to arise.

ky28059: Ex.
```java
// Periodic loop
if (mech.getRightBumperPressed()) {
   elevator.height = elevator.height.next();
}
if (mech.getLeftBumperPressed()) {
    elevator.height = elevator.height.previous();
}
```
```
// Button callbacks in `configureButtonBindings()`
// Made even cleaner with methods and method references (ie. `elevator::nextPosition` or `elevator::previousPosition`)
mechRightBumper.whenPressed(new InstantCommand(() -> elevator.height = elevator.height.next()));
mechLeftBumper.whenPressed(new InstantCommand(() -> elevator.height = elevator.height.previous()));
```

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -25,27 +28,16 @@
  */
 public class RobotContainer {
     // Subsystems
-    //private final SwerveSubsystem swerveSubsystem;
-    private final MissileShellSwerveSubsystem swerveSubsystem;
+    // private final SwerveSubsystem swerveSubsystem;
+    final GripperSubsystem gripper = new GripperSubsystem();
```
ky28059, line 32: Package private 😳

ky28059: Deleting the boilerplate 😔 

image

PR #15, opened 2022-11-25
7th Shop Project

e3l:
PR for comments


src/main/java/frc/robot/Constants.java
```diff
@@ -17,28 +16,37 @@
...
+        public static final int RIGHT_MAIN = 3;
+        public static final int RIGHT_SECONDARY = 4;
+
+    }
+    public static final class IntakeConstants{
+        public static final int FRONT_MOTOR = 5;
+        public static final int RIGHT_MOTOR = 6;
+        public static final int LEFT_MOTOR = 7;
+
+        public static final int INTAKE_SERVO = 10; //servo pin ID
```
e3l, line 32: the PWM ports on the RIO are #'d 0-9, you can probably just assign ports now

src/main/java/frc/robot/Constants.java
```diff
@@ -17,28 +16,37 @@
...
+        public static final int INTAKE_SERVO_DOWN = 90; //degree value
+    }
+    public static final class CarriageConstants{
+        public static final int CARRIAGE_SOLENOID = 1;
+        public static final int TOP_SERVO = 1;
+        public static final int BOTTOM_SERVO = 1;
+
+        public static final int TOP_OPEN = 90; // servo position when the door is open
```
e3l, line 40: make sure to read the docs on the range of inputs to setting a Servo

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -89,4 +75,26 @@ private void configureButtonBindings() {
...
+
+    void periodic(){
+        tank.forwardComponent = -1 * controller.getLeftY(); // 1 is forward (adjusted from -1 forward)
+        tank.sideComponent = controller.getRightX(); // 1 is right
+
+        if(controller.getBButton()){ //engage the intake arm by pressing B on the controller
+            intake.intake_down = true;
+
+        }
+        if(controller.getXButton()){
```
e3l, line 87: be wary that this method will return true every loop that the button is held... so if the button is held for more than a split second (whether intentionally or not), this might call more times than you'd like

src/main/java/frc/robot/subsystems/Intake.java
```diff
@@ -0,0 +1,75 @@
...
+public class Intake extends SubsystemBase {
+    
+    private final CANSparkMax front = new CANSparkMax(FRONT_MOTOR, MotorType.kBrushless); // front motor for intake
+    private final WPI_TalonSRX right = new WPI_TalonSRX​(RIGHT_MOTOR); // right roller between intake and carriage
+    private final WPI_TalonSRX left = new WPI_TalonSRX​(LEFT_MOTOR); // left roller between intake and carriage
+
+    Servo intake_servo = new Servo(INTAKE_SERVO);
+
+    public boolean intake_down = false;
+    public boolean intake_on = false;
```
e3l, line 34: i think we should put states at the very top for readability

also, i think drivers would appreciate being able to reverse the intake or go slower, which is why we usually put the intake on (two) triggers

src/main/java/frc/robot/subsystems/Tank.java
```diff
@@ -0,0 +1,71 @@
...
+
+    leftDrive = forwardComponent + sideComponent;
+    rightDrive = forwardComponent - sideComponent;
+
+    if(Math.abs(leftDrive) >= 1.0){
+        leftDrive = leftDrive / Math.abs(leftDrive);
+    }
+    if(Math.abs(rightDrive) >= 1.0){
+        rightDrive = rightDrive / Math.abs(rightDrive);
+    }
```
e3l, line 60: typically we scale both powers by the same factor so that differences between left/right aren't just cut off, for example if left power is 0.5 and right is 1.5. If we want to keep turn sensitivity, just cutting that to 0.5 and 1.0 isn't great

e3l:
also, it seems like your code isn't building (via Github Actions and also on my laptop) because there are some really funky characters (hidden?) in Tank.java. Does it build on your laptop locally?

src/main/java/frc/robot/subsystems/Tank.java
```diff
@@ -0,0 +1,71 @@
...
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX; // internal Falcon motors
+import com.ctre.phoenix.motorcontrol.NeutralMode;
+
+import edu.wpi.first.wpilibj.XboxController; // controller input library
+
+import static frc.robot.Constants.TankConstants.*;
+
+public class Tank extends SubsystemBase {
+    
+    private final WPI_TalonSRX left = new WPI_TalonSRX​(LEFT_MAIN); // left motor
```
ky28059, line 19: You seem to have [zero-width spaces](https://en.wikipedia.org/wiki/Zero-width_space) `ZWSP` sprinkled among your `WPI_TalonSRX` constructor calls?

image

Removing these should fix the build errors (I think VSCode should render these characters as boxes with `ZWSP` inside them? If not it may be a setting you have to enable). 

Also, I'd recommend using `MotorUtil.createTalonSRX()` for easier global configuration.

src/main/java/frc/robot/subsystems/Tank.java
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) FIRST and other WPILib contributors.
+// Open Source Software; you can modify and/or share it under the terms of
+// the WPILib BSD license file in the root directory of this project.
+
+package frc.robot.subsystems;
+
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+import com.ctre.phoenix.motorcontrol.can.BaseMotorController; 
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX; // internal Falcon motors
```
ky28059, line 10: These are not Falcon motors; the `TalonSRX`s on the base are connected to CIM motors. You may be thinking of `TalonFX`, which is the integrated motor controller inside Falcon motors, but we do not currently have or use any Falcons.

src/main/java/frc/robot/subsystems/Intake.java
```diff
@@ -0,0 +1,75 @@
...
+import com.revrobotics.CANSparkMax.IdleMode;
+
+import edu.wpi.first.wpilibj.Servo;
+
+import static frc.robot.Constants.IntakeConstants.*;
+
+
+public class Intake extends SubsystemBase {
+    
+    private final CANSparkMax front = new CANSparkMax(FRONT_MOTOR, MotorType.kBrushless); // front motor for intake
```
ky28059, line 27: Use `MotorUtil.createSparkMax(int)`? Same with the other `TalonSRX` constructors; using the util lets you not have to worry about factory resetting, since the method automatically does that and applies global config settings.

src/main/java/frc/robot/subsystems/Carriage.java
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) FIRST and other WPILib contributors.
+// Open Source Software; you can modify and/or share it under the terms of
+// the WPILib BSD license file in the root directory of this project.
+
+package frc.robot.subsystems;
+
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+import com.ctre.phoenix.motorcontrol.can.BaseMotorController; 
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX; // internal Falcon motors
```
ky28059, line 10: See comment below.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -89,4 +75,26 @@ private void configureButtonBindings() {
     public Command getAutonomousCommand() {
         return autonChooser.getSelected();
     }
+
+    void periodic(){
+        tank.forwardComponent = -1 * controller.getLeftY(); // 1 is forward (adjusted from -1 forward)
+        tank.sideComponent = controller.getRightX(); // 1 is right
+
+        if(controller.getBButton()){ //engage the intake arm by pressing B on the controller
```
ky28059, line 83: On top of what ethan said, you can implement this logic with callback-based `JoystickButton`s instead of having it in periodic. The `controller.getXButton()` can be rewritten as `controllerXButton.whenPressed(...)` to run a command that toggles `intake_on` when the button is pressed, and `controller.getBButton()` can be rewritten as `controllerBButton.whileHeld(...)` (unless you want it to be toggle-on-press like the X button).

It's a little more complicated since it uses the command system and callbacks, but I think it makes it cleaner (and handles debouncing and errors from toggling in periodic like below).

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -25,27 +31,13 @@
...
-
-    private final XboxController mechController = new XboxController(1);
-    private final JoystickButton 
-        mechAButton = new JoystickButton(mechController, XboxController.Button.kA.value),
-        mechBButton = new JoystickButton(mechController, XboxController.Button.kB.value),
-        mechXButton = new JoystickButton(mechController, XboxController.Button.kX.value),
-        mechYButton = new JoystickButton(mechController, XboxController.Button.kY.value),
-        mechLBumper = new JoystickButton(mechController, XboxController.Button.kLeftBumper.value),
-        mechRBumper = new JoystickButton(mechController, XboxController.Button.kRightBumper.value);
+    public final XboxController controller = new XboxController(0);
```
ky28059, line 39: You'll end up having two controllers anyways (mech and drive), so technically this boilerplate that constructs both controllers and all the buttons didn't have to be removed.

ky28059 (changes requested):

src/main/java/frc/robot/subsystems/Tank.java
```diff
@@ -41,6 +41,8 @@ public Tank() {
     left.setNeutralMode(NeutralMode.Brake);
     right.setNeutralMode(NeutralMode.Brake);
     right.setInverted(true);
+    right2.setInverted(true);
```
ky28059, line 44: See https://github.com/grt192/GRT2023/pull/14#discussion_r1038901131.

e3l:
make sure to write new software for the changed intake too :')

src/main/java/frc/robot/subsystems/Carriage.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Override
+  public void periodic() {
+    // This method will be called once per scheduler run
+
+    
+    if(liftCarriage){
+      piston.toggle();
+      liftCarriage = false;
```
e3l, line 45: if the state directly represents where we want to carriage to be, why not just set the piston based on that? seems weird to modify the state after achieving it

src/main/java/frc/robot/subsystems/Intake.java
e3l: based on our intake tests, full power is a little too fast, so we'll definitely want to be able to set a variable power as the state. 

We'll proably even want to scale that power before setting the state in RobotContainer so full trigger sets like 0.5 power

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -89,4 +84,31 @@ private void configureButtonBindings() {
     public Command getAutonomousCommand() {
         return autonChooser.getSelected();
     }
+
+    void periodic(){
+        tank.forwardComponent = -1 * driveController.getLeftY(); // 1 is forward (adjusted from -1 forward)
+        tank.sideComponent = driveController.getRightX(); // 1 is right
+        
+        if(driveController.getRightBumperPressed() && driveController.getLeftBumperPressed()){
```
ky28059, line 92: See https://github.com/grt192/GRT2023/pull/14#discussion_r1048086571.

src/main/java/frc/robot/subsystems/Carriage.java
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) FIRST and other WPILib contributors.
+// Open Source Software; you can modify and/or share it under the terms of
+// the WPILib BSD license file in the root directory of this project.
+
+package frc.robot.subsystems;
+
+import static frc.robot.Constants.CarriageConstants.BOTTOM_CLOSED;
```
ky28059, line 7:
```java
import static frc.robot.Constants.CarriageConstants.*;
```
?

src/main/java/frc/robot/subsystems/Carriage.java
```diff
@@ -0,0 +1,88 @@
...
+       
+      if(!timer_enabled){ //if the timer was not enabled, close the door and then start the timer
+        top.setAngle(TOP_CLOSED);
+        bottom.setAngle(BOTTOM_CLOSED);
+        openDoor = false;
+        timer.reset(); // reset and start timer to make sure doors are closed before lift
+        timer.start();
+        timer_enabled = true;
+      }
+      if(timer.get() >= 0.2){ // if 0.2 seconds pass, set the piston forward
```
ky28059, line 58:
```java
timer.hasElapsed(0.2)
```
or
```java
timer.advanceIfElapsed(0.2)
```
if you want to do something on a precise loop.

src/main/java/frc/robot/subsystems/Intake.java
```diff
@@ -0,0 +1,80 @@
...
+    if(intake_power_forward <= 0.10){ // if forward intake trigger is not pressed significantly, dont make it spin
+      intake_power += 0.0;
+    }
+    if(intake_power_forward > 0.10 && intake_power_forward < 0.85){ // if its pressed a little, make it spin slow
+      intake_power += 0.18;
+    }
+    if(intake_power_forward > 0.85){
+      intake_power += 0.35; // if its pressed a lot, make it spin fast
+    }
+    if(intake_power_reverse >= 0.4){ // if there is some reverse component, make it spin that way
```
ky28059, line 67: Isn't this chain of `if`s roughly equivalent to
```java
intake_power = intake_power_forward * 0.35 - intake_power_reverse * 0.45;
```
? If you want to make it unresponsive to low inputs you can apply a [deadband](https://team2168.org/javadoc/wpilib/edu/wpi/first/math/MathUtil.html#applyDeadband(double,double)) to the joystick values. Also unsure what the value is to scaling the forward and reverse powers differently? You could probably just do
```java
intake_power = (intake_power_forward - intake_power_reverse) * 0.45;
```
like we did with the intake last year.

src/main/java/frc/robot/subsystems/Tank.java
```diff
@@ -0,0 +1,79 @@
...
+    private final WPI_TalonSRX left = MotorUtil.createTalonSRX(LEFT_MAIN); // left motor
+    private final WPI_TalonSRX left2 = MotorUtil.createTalonSRX(LEFT_SECONDARY); // left motor
+
+    private final WPI_TalonSRX right = MotorUtil.createTalonSRX(RIGHT_MAIN); // right motor
+    private final WPI_TalonSRX right2 = MotorUtil.createTalonSRX(RIGHT_SECONDARY); // right motor
+
+    public double sideComponent;
+    public double forwardComponent;
+  
+    double leftDrive;
```
ky28059, line 31: These don't need to be stored as fields?

PR #20, opened 2023-01-10
Use PhotonVision

ky28059:


src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -16,9 +16,13 @@
 import edu.wpi.first.wpilibj.Timer;
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
 
+import frc.robot.vision.PhotonVision;
+
 import static frc.robot.Constants.SwerveConstants.*;
 
 public class SwerveSubsystem extends SubsystemBase {
+    private final PhotonVision vision;
```
e3l, line 24: small detail, but perhaps this class should be called "PhotonVisionConnection" or something for clarity?

src/main/java/frc/robot/subsystems/MissileShellVisionSubsystem.java
```diff
@@ -0,0 +1,31 @@
...
+        this.vision = vision;
+    }
+
+    @Override
+    public void periodic() {
+        var res = vision.getRobotPose(pose);
+        this.pose = res.getFirst();
+
+        xEntry.setDouble(pose.getX());
+        yEntry.setDouble(pose.getY());
```
e3l, line 29: oo shuffleboard is pretty easy

let's add the robot angle too?

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -125,11 +131,15 @@ public SwerveDriveKinematics getKinematics() {
...
         // Update pose estimator from swerve module states
         Rotation2d gyroAngle = getGyroHeading();
-        poseEstimator.update(
+        Pose2d estimate = poseEstimator.update(
             gyroAngle,
             getModuleStates()
         );
 
+        // Add vision pose estimate to pose estimator
+        var res = vision.getRobotPose(estimate);
```
e3l, line 140: why 'var'? why not just write the type :( i can't see inferences on web github

is this a blocking call? does this wait for communications over networking? if so, we shouldn't tie up periodic waiting for this to happen

src/main/java/frc/robot/vision/PhotonVision.java
```diff
@@ -0,0 +1,70 @@
...
+    private final RobotPoseEstimator visionPoseEstimator;
+
+    /**
+     * Constructs a PhotonVision connection to the coprocessor.
+     */
+    public PhotonVision() {
+        camera = new PhotonCamera(cameraName);
+
+        // Create list of cameras and camera positions
+        var camList = new ArrayList<Pair<PhotonCamera, Transform3d>>();
```
e3l, line 36: can we maybe generate this list from an array in constants? so we can just add an element to an array in Constants.java to add more cameras

src/main/java/frc/robot/subsystems/MissileShellVisionSubsystem.java
ky28059: Perhaps? I deleted the test class entirely because I wasn't sure if the info it was getting was valuable at all. The vision pose estimator needs a previous reference pose to work, but it seems like passing its old estimate back to it creates some sort of feedback loop where it just gives the same position over and over? Either that or there was something off with the camera position or how it was mounted; I think it might be easier to mount it on an actual swerve base with odometry to test.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
ky28059: I used `var` because I didn't want to write `Pair<Pose2d, Double>` :D the example code I copied from PhotonVision also uses `var` for some nasty types (like I think the pose estimator itself returns like `Optional<Pair<Pose3d, Double>>` or smth); it's also as far as I can tell not a blocking call in the same way the odometry pose estimator update is not a blocking call.

src/main/java/frc/robot/vision/PhotonVision.java
ky28059: Good idea 👍

src/main/java/frc/robot/vision/PhotonVision.java
```diff
@@ -0,0 +1,70 @@
...
+        camList.add(new Pair<PhotonCamera, Transform3d>(camera, cameraPos));
+
+        // TODO: is this hacky?
+        try {
+            visionPoseEstimator = new RobotPoseEstimator(
+                new AprilTagFieldLayout(AprilTagFields.k2022RapidReact.m_resourceFile), // TODO: update for 2023 charged up
+                PoseStrategy.CLOSEST_TO_REFERENCE_POSE,
```
e3l, line 43: would we have to test the ambiguity and best average settings? they seem promising

src/main/java/frc/robot/subsystems/MissileShellVisionSubsystem.java
e3l: this seems really critical to debug even before running on an actual swerve base. 
are vision estimates currently passed to the poseestimator even if they're null? what's the behavior of doing so

src/main/java/frc/robot/Constants.java
```diff
@@ -52,6 +55,15 @@ public static final class SwerveConstants {
         );
     }
 
+    // TODO
```
e3l, line 58: also please document what system the translation/rotations are in?

ky28059:
Superseded by #33.

PR #21, opened 2023-01-11
Kopbackup

rishay-jain:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -92,14 +100,31 @@ public RobotContainer() {
...
             double xPower = -driveController.getLeftY();
             double yPower = -driveController.getLeftX();
             double angularPower = -driveController.getRightX();
             swerveSubsystem.setSwerveDrivePowers(xPower, yPower, angularPower);
         }, swerveSubsystem));
+        */
     }
 
+
+    void periodic(){
```
e3l, line 114: ideally we don't have to have a periodic method here, it would be best if it can be a defaultCommand like it was for swerve.

I wonder if it's possible to have both swerve and tank classes extend a common Drivetrain class. then we can just have an if statement depending on whether the current drivetrain extends Swerve or extends Tank, and set the default command of the drivetrain correspondingly

src/main/java/frc/robot/subsystems/Balancer.java
```diff
@@ -0,0 +1,84 @@
+package frc.robot.subsystems;
+import com.kauailabs.navx.frc.AHRS;
+import edu.wpi.first.wpilibj.SPI;
+
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class Balancer extends SubsystemBase {
```
e3l, line 7: i think it's more inline with the intent of the command based drivestructure to do this using a Command. Is there a way to cleanly implement this using a command? 

that way the default command will also be overriden by the balancecommand

src/main/java/frc/robot/subsystems/Tank.java
```diff
@@ -0,0 +1,75 @@
...
+
+public class Tank extends SubsystemBase {
+    
+    public boolean driveInvert;
+
+    private final WPI_TalonSRX left = MotorUtil.createTalonSRX(LEFT_MAIN); // left motor
+    private final WPI_TalonSRX left2 = MotorUtil.createTalonSRX(LEFT_SECONDARY); // left motor
+
+    private final WPI_TalonSRX right = MotorUtil.createTalonSRX(RIGHT_MAIN); // right motor
+    private final WPI_TalonSRX right2 = MotorUtil.createTalonSRX(RIGHT_SECONDARY); // right motor
```
e3l, line 22: these comments are riveting, insightful, and eye-opening

src/main/java/frc/robot/subsystems/Tank.java
```diff
@@ -0,0 +1,75 @@
...
+
+  }
+
+
+  @Override
+  public void periodic() {
+    // This method will be called once per scheduler run
+
+    leftDrive = forwardComponent + sideComponent*0.75;
+    rightDrive = forwardComponent - sideComponent*0.75;
```
e3l, line 49: seems like an odd thing to limit in the subsystem code, rather than in RobotContainer. We shouldn't prevent autonomous from being able to use the full range of turnpower, for example.

src/main/java/frc/robot/commands/BalancerCommand.java
```diff
@@ -0,0 +1,44 @@
...
+  private final Balancer m_autobal;
+  public double returnPower;
+
+  /**
+   * Creates a new ExampleCommand.
+   *
+   * @param subsystem The subsystem used by this command.
+   */
+  public BalancerCommand(Balancer subsystem) {
+    m_autobal = subsystem;
```
e3l, line 18: so i think the intent of this command-based structure is for the subsystem to be the drivetrain, and all the logic for the balancer goes in the execute() method (which gets called similarly to the periodic() method of a subsystem). so essentially we do not need a balancer subsystem, which makes sense because there is no physical "balancer mech"

src/main/java/frc/robot/subsystems/Balancer.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public int phase = 1;
```
e3l, line 9: usually it's clearer to use java ENUMS to represent states, instead of ints, for readability

src/main/java/frc/robot/commands/BalancerCommand.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {}
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    return false;
```
e3l, line 42: this command should say that it's done at some point

src/main/java/frc/robot/commands/BalancerCommand.java
```diff
@@ -0,0 +1,106 @@
...
+        default:
+            break;
+    }
+    if(dtType == "swerve"){
+        swerveSubsystem.setSwerveDrivePowers(returnPower,0.0,initialHeading);    
+    }
+    if(dtType == "tank"){
+        tank.forwardComponent = returnPower;
+        tank.sideComponent = 0.0;
+    }
```
e3l, line 94: this is cool but it could be even cooler if both Tank and Swerve extended some abstract Drivetrain class that had a method like `.setPower(double yPower, double turnPower)`. That way there can just be one field `Drivetrain dt` for this class and `dt.setPower(returnPower,0)` would work no matter if the DT passed is swerve or tank

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -26,13 +31,17 @@
  */
 public class RobotContainer {
     // Subsystems
-    //private final SwerveSubsystem swerveSubsystem;
-    private final MissileShellSwerveSubsystem swerveSubsystem;
+    private final Tank dt; // declare DT of choice
```
e3l, line 34: should be Drivetrain instead of Tank, right?

src/main/java/frc/robot/subsystems/drivetrain/SwerveSubsystem.java
```diff
@@ -81,7 +82,7 @@ public SwerveSubsystem() {
      * @param yPower The power [-1.0, 1.0] in the y (left) direction.
      * @param angularPower The angular (rotational) power [-1.0, 1.0].
      */
-    public void setSwerveDrivePowers(double xPower, double yPower, double angularPower) {
+    public void setDTDrivePowers() {
```
e3l, line 85: this is not the same method defined in DriveTrain.java

e3l (changes requested):

src/main/java/frc/robot/Robot.java
```diff
@@ -62,7 +62,9 @@ public void teleopInit() {
     }
 
     @Override
-    public void teleopPeriodic() {}
+    public void teleopPeriodic() {
+        robotContainer.periodic();
```
e3l, line 66: can be removed now?

src/main/java/frc/robot/subsystems/drivetrain/DriveTrain.java
```diff
@@ -0,0 +1,18 @@
...
+
+    double xPower;
+    double yPower;
+    double angularPower;
+
+    public void updateDrivePowers(double forwardComponent, double sideComponent, double angularPower){
+        // xPower = forwardComponent;
+        // yPower = sideComponent;
+        // this.angularPower = angularPower;
+        // System.out.println("dt received drive powers");
```
e3l, line 15: because we don't need (and in fact, don't want) to define a default `updateDrivePowers` method, this class can either be an `abstract class` or an `interface`.

src/main/java/frc/robot/subsystems/drivetrain/DriveTrain.java
```diff
@@ -0,0 +1,18 @@
+package frc.robot.subsystems.drivetrain;
+
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class DriveTrain extends SubsystemBase {
```
e3l, line 5: very tiny thing but i dislike how this is DriveTrain instead of Drivetrain

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -67,24 +79,28 @@ public class RobotContainer {
...
-
+    private final Command balanceCommand;
     /**
      * The container for the robot. Contains subsystems, OI devices, and commands.
      */
     public RobotContainer() {
-        swerveSubsystem = new SwerveSubsystem2020();
-        // swerveSubsystem = new MissileShellSwerveSubsystem();
+        dt = new Tank(); // initialize DT of choice
+        ahrs = new AHRS(SPI.Port.kMXP);
```
e3l, line 88: i don't think we should be constructing multiple ahrs, i don't know what the behavior is but I doubt it's good. Maybe we can have the AHRS in the Drivetrain class, and offer methods to get data from the ahrs? so the balancecommand doesn't need the ahrs either

e3l (approved):
some formatting (indenting) is really wonky, and I think BalanceCommand could access the NavX directly instead of being passed it as a parameter. Otherwise lgtm as soon as it builds!

src/main/java/frc/robot/commands/BalancerCommand.java
```diff
@@ -0,0 +1,117 @@
...
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
+import frc.robot.subsystems.drivetrain.Drivetrain;
+import frc.robot.subsystems.drivetrain.Tank;
+
+/** An example command that uses an example subsystem. */
+public class BalancerCommand extends CommandBase {
+
+  private Drivetrain driveTrain;
+  String dtType;
```
e3l, line 16: let's clean up unused field variables pls

src/main/java/frc/robot/subsystems/drivetrain/DriveTrain.java
```diff
@@ -0,0 +1,17 @@
...
+
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import com.kauailabs.navx.frc.AHRS;
+import edu.wpi.first.wpilibj.SPI;
+
+public class Drivetrain extends SubsystemBase {
+    
+    public final AHRS ahrs;
+
+    public Drivetrain(){
```
e3l, line 11: hm i just realized, can't an abstract class still have a constructor? this way updateDrivePowers could be an abstract method

src/main/java/frc/robot/subsystems/drivetrain/Tank.java
```diff
@@ -0,0 +1,98 @@
...
+    right2.follow(right);
+
+    left.setNeutralMode(NeutralMode.Brake);
+    right.setNeutralMode(NeutralMode.Brake);
+    right.setInverted(true);
+    right2.setInverted(true);
+
+    
+
+
```
e3l, line 41: so many empty lines :(

src/main/java/frc/robot/subsystems/drivetrain/Tank.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  public void updateDrivePowers(double forwardComponent, double sideComponent, double angularPower){
+    // POSSIBLE/CONCEPT -- double theta = Math.atan(yPower/xPower); // if the robot needs to move some amount laterally, it can rotate to this angle and move in that direction
+    
+    leftDrive = xPower + yPower;
+    rightDrive = xPower - yPower;
```
e3l, line 48: if these are only used in the scope of this method, it makes sense to me to declare them here (rather than as field variables)

e3l (commented):
actually some more comments too

PR #22, opened 2023-01-11
Adds ability to reset field angle

penguin212:
The resetFieldAngle function resets the forward position for the joystick when it is called (not assigned to a button yet).


src/main/java/frc/robot/subsystems/swerve/SwerveModule2020.java
```diff
@@ -0,0 +1,147 @@
...
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.kinematics.SwerveModuleState;
+
+import frc.robot.motorcontrol.MotorUtil;
+
+/**
+ * A swerve module with a NEO drive motor and a BAG steer motor, for running swerve modules on 
+ * the 2020 robot.
+ */
+public class SwerveModule2020 {
```
e3l, line 24: wait is this class duplicated

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -26,8 +28,8 @@
  */
 public class RobotContainer {
     // Subsystems
-    //private final SwerveSubsystem swerveSubsystem;
-    private final MissileShellSwerveSubsystem swerveSubsystem;
+    private final ISwerveSubsystem swerveSubsystem;
```
e3l, line 31: small detail... i don't really like the "I" naming lol, esp cuz it's not an interface... what about AbstractSwerveSubsystem, perhaps

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
```diff
@@ -211,4 +214,13 @@ public void resetPosition() {
     private Rotation2d getGyroHeading() {
         return Rotation2d.fromDegrees(-ahrs.getAngle());
     }
+
+    private Rotation2d getGyroForField() {
+        return Rotation2d.fromDegrees(-(ahrs.getAngle() - angleoffset));
+    }
+
+    public void resetFieldAngle() {
```
e3l, line 222: the final challenge is abstracting this code (that's common between the two years) too :))))

src/main/java/frc/robot/RobotContainer.java
ky28059: Yeah, as an abstract class `BaseSwerveSubsystem` makes more sense.

src/main/java/frc/robot/subsystems/swerve/SwerveSubsystem.java
ky28059: The states, gyro, and `setSwerveDrivePowers` can be abstracted too! Maybe even periodic? The hardest part would be abstracting the modules and adding odometry to 2020.

src/main/java/frc/robot/subsystems/swerve/SwerveModule2020.java
ky28059: The difference between the 2020 swerve module and 2023 is that 2020 uses different motors and needs different PID constants to run. Also 2020 doesn't have the subclass setup in place yet, which we could consider adding.

src/main/java/frc/robot/subsystems/swerve/SwerveModule.java
```diff
@@ -115,29 +118,32 @@ public SwerveModule(int drivePort, int steerPort) {
...
      */
     public SwerveModulePosition getState() {
         return new SwerveModulePosition(
-            // driveMotor.getSelectedSensorPosition() * DRIVE_TICKS_TO_METERS,
-            driveEncoder.getPosition(),
-            getAngle()
-        );
+                // driveMotor.getSelectedSensorPosition() * DRIVE_TICKS_TO_METERS,
+                driveEncoder.getPosition(),
+                getAngle());
```
ky28059, line 128: Oh you used format file didn't you :( now it'll make multi-line constructors and method invocations look ugly like this and break lines suboptimally

e3l (approved):
would recommend renaming `ISwerveSubsystem` to `AbstractSwerveSubsystem` or something similar.

Otherwise lgtm, can merge if you don't plan on more aggressively abstracting

ky28059 (approved):
👍

e3l (approved):
lgtm

PR #23, opened 2023-01-13
Robot 0 Grippers

CrolineCrois:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -107,4 +107,16 @@ private void configureButtonBindings() {
...
+
+    public void periodic(){
+        //check if button presses, switch state accordingly
+        if (mechController.getAButtonPressed() == true){
+            if (gripper.state == Value.kForward){
+                gripper.state = Value.kReverse;
+            }
+            else {
+                gripper.state = Value.kForward;
+            }
```
e3l, line 119: try using commands instead of making a periodic: see https://github.com/grt192/GRT2022/blob/b6262bdcf0f6f2c0bc4704c04ddbf55af00cbb7f/src/main/java/frc/robot/RobotContainer.java#L169 for an example. i think you could make a toggleClamp in the subsystem and make an instant command with gripper::toggleClamp?

src/main/java/frc/robot/subsystems/GripperSubsytem.java
```diff
@@ -0,0 +1,34 @@
...
+
+    DoubleSolenoid pfft = new DoubleSolenoid(PneumaticsModuleType.CTREPCM, 0, 1);
+
+    public GripperSubsytem(){
+        //first set to forward
+        state = Value.kForward;
+    }
+
+    //instant command method
+    public void GripToggle(){
```
e3l, line 20: methods should be lowerCamelCase :D

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -0,0 +1,68 @@
...
+    
+    private final WPI_TalonSRX leftbeak = MotorUtil.createTalonSRX(LEFTID);
+    private final WPI_TalonSRX rightbeak = MotorUtil.createTalonSRX(RIGHTID);
+    
+    //limit switch
+    private final DigitalInput crimitswitch = new DigitalInput(0);
+    //timer
+    public Timer timere = new Timer();
+    //roller state 
+    public String rollstate;
```
e3l, line 23: if we use a Java Enum instead we can make sure the value is one of the few that we expect.

that being said, don't we want to give drivers the option to control how fast it's intaking? i think having a double intakePower as this state wouldn't be too unreasonable

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public void periodic() {
+        //if wheels must intake, and the limit switch is not pressed, turn on motors
+        if (rollstate == "in" && !crimitswitch.get()){
+            leftbeak.set(.5);
+        }
+        //if wheels must place, turn on timer then turn on motors for 2 seconds, reset timer
+        else if (rollstate == "out"){
+            timere.start();
+            while (timere.get() < 2000){
```
e3l, line 50: no while loops in periodic! this'll freeze all the rest of our code. this logic should check if it's been 2000 seconds every time it's called

also, i think this delay should be a constant field variable, and we can probably turn it to a really high # if we decide we can just leave it to the drivers to choose when to stop outtaking. i don't think there's really a harm from running the outtake too long?

e3l (changes requested):
also be sure to rebase onto main after we merge rishay's pull request

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -119,6 +128,22 @@ private void configureButtonBindings() {
                 tankSubsystem.setDrivePowers(forwardPower, turnPower);
             }, tankSubsystem));
         }
+
+        roller.setDefaultCommand(new RunCommand(() -> {
+            if (mechController.getRightTriggerAxis() > .2){
+                roller.rollstate = mechController.getRightTriggerAxis();
+                System.out.println("b");
```
e3l, line 135: can remove

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -119,6 +128,22 @@ private void configureButtonBindings() {
...
+            }
+            else if (mechController.getLeftTriggerAxis() > .2){
+                roller.rollstate = -(mechController.getLeftTriggerAxis());
+            }
+            else{
+                roller.rollstate = 0.0;
+            }
```
e3l, line 142: instead of if statements we can just add the values from both triggers i think

src/main/java/frc/robot/subsystems/GripperSubsytem.java
```diff
@@ -0,0 +1,36 @@
...
+            state = Value.kForward;
+        }
+    }
+
+    @Override
+    public void periodic() {
+        pfft.set(state);
+        System.out.println(state);
+    }
+     }
```
e3l, line 36: indentation goes crazy

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -0,0 +1,74 @@
...
+            
+        }
+        
+    }
+
+
+
+
+
+
```
e3l, line 73: can clean up formatting on this doc and remove extra system.prints. otherwise i think its good

PR #25, opened 2023-01-14
Shira mech controls, offsets, base soft limits

penguin212:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -101,7 +107,36 @@ private void configureButtonBindings() {
...
+
+        mechRBumper.onTrue(new InstantCommand(() ->{
+            if(moverSubsystem.getState() == MoverPosition.CUBEHIGH){
+                moverSubsystem.setState(MoverPosition.CUBEMID);
+            } else {
+                moverSubsystem.setState(MoverPosition.CUBEHIGH);
+            }
+        }, moverSubsystem));
+
+        mechLBumper.onTrue(new InstantCommand(() ->{
```
e3l, line 133: will onTrue cause this logic to be run thousands of times each time the button is held down? i recall we used to use whenPressed or something

src/main/java/frc/robot/subsystems/MoverSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+    
+    public enum MoverPosition {
+        VERTICAL(0,0),
+        GROUND(Units.degreesToRadians(90), distanceToExtension(25)),
+        SUBSTATION(0,0),
+        CUBEMID(0,0), CUBEHIGH(0,0), 
+        CONEMID(0,0), CONEHIGH(0,0)
+        ;
+
+        public double angle; //Radians
```
e3l, line 53: pls also comment where 0 is

src/main/java/frc/robot/subsystems/MoverSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+import com.revrobotics.CANSparkMax.SoftLimitDirection;
+
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.motorcontrol.MotorUtil;
+import static frc.robot.Constants.MoverConstants.*;
+
+public class MoverSubsystem extends SubsystemBase{
+
+    private final CANSparkMax rotationMotor;
```
e3l, line 17: there are two of these in the gearbox, i think

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -101,7 +107,36 @@ private void configureButtonBindings() {
             swerveSubsystem.setSwerveDrivePowers(xPower, yPower, angularPower);
         }, swerveSubsystem));
 
+        moverSubsystem.setDefaultCommand(new RunCommand(() -> {
+            double xPower = -mechController.getLeftX();
+            double yPower = -mechController.getRightY();
```
e3l, line 112: this command doesn't seem to actually do anything besides save values to booleans

maybe for just seeing motor directions we can just make a simple subsystem to use temporarily

src/main/java/frc/robot/subsystems/MoverSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+    public MoverSubsystem(){
+        rotationMotor = MotorUtil.createSparkMax(ROTATION_MOTOR_PORT);
+        rotationMotor.setIdleMode(IdleMode.kBrake);
+
+        rotationEncoder = rotationMotor.getEncoder();
+        rotationEncoder.setPositionConversionFactor(ROTATION_ROT_TO_RAD);
+        
+        rotationMotor.enableSoftLimit(SoftLimitDirection.kForward, true);
+        rotationMotor.enableSoftLimit(SoftLimitDirection.kReverse, true);
+        rotationMotor.setSoftLimit(SoftLimitDirection.kForward, (float) Units.degreesToRadians(90));
```
e3l, line 71: won't it take more than 90 degrees to get to the ground? i'm guessing 0 degrees is straight up

src/main/java/frc/robot/subsystems/MoverSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Override
+    public void periodic(){
+        goTo(currentState.angle + angleOffset, currentState.extension + extensionOffset);
+    }
+
+
+    private void goTo(double angle, double extension){
+        rotationPidController.setReference(angle, ControlType.kPosition);
+        extensionPidController.setReference(extension, ControlType.kPosition);
```
e3l, line 106: it will really help to put data on target/current position on shuffleboard, along with being able to tune the PID constants there. otherwise it will take a long time to wait for reupload every time.

e3l (changes requested):

src/main/java/frc/robot/RobotContainer.java
penguin212: The old function used in 2022 was marked as deprecated, onTrue just catched when it switches from false to true. I think whileTrue continuously calls it.

src/main/java/frc/robot/subsystems/MoverSubsystem.java
penguin212: I think it will depend on the grabber, we can definitely change that later. I just didn't want to accidently slam into the ground on the first test.

src/main/java/frc/robot/Constants.java
```diff
@@ -93,4 +93,20 @@ public static final class SwerveConstants2020 {
             Units.inchesToMeters(-10.3865)
         );
     }
+
+    public static final class ShuffleboardConstants {
+        public static final double UPDATE_TIME = 0.5; // seconds between each update
```
e3l, line 98: this isn't relevant for the new shuffleboard, right

e3l (approved):

PR #26, opened 2023-01-14
Swerve robot-relative control

ky28059:
Need to test on 2020 robot, but otherwise seems mergeable?

### DIFF
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -97,19 +97,27 @@ public class RobotContainer {
     private void configureButtonBindings() {
         driveRBumper.whileTrue(balancerCommand);
 
-        if (driveSubsystem instanceof BaseSwerveSubsystem) {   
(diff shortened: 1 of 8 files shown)


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -98,7 +98,8 @@ private void configureButtonBindings() {
             double xPower = -driveController.getLeftY();
             double yPower = -driveController.getLeftX();
             double angularPower = -driveController.getRightX();
-            swerveSubsystem.setSwerveDrivePowers(xPower, yPower, angularPower);
+            boolean relative = driveController.getRightTriggerAxis() > 0.75;
+            swerveSubsystem.setSwerveDrivePowers(xPower, yPower, angularPower, relative);
```
e3l, line 102: this'll be pretty awkward with rishay's abstractions.. perhaps in this section we can just use an if to cast to Tank or Swerve and just have all the setting logic separate.

e3l (approved):
lgtm, i think should be rebased on main after rishay's PR and we should test this before merging.

e3l (approved):
sweet

PR #29, opened 2023-01-15
Fall back on pose estimator for field-centric heading

ky28059:
Currently this PR keeps the current system where resetting the field angle has no effect on localization. If resetting the field angle should also affect localization, we can either make resetting the field angle reset the localization to (0,0) or maybe try to guess the robot's new position with the angle changed (but this is probably hacky and inconsistent).

### DIFF
--- a/src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
@@ -224,9 +224,17 @@ public abstract class BaseSwerveSubsystem extends BaseDrivetrain {
 
     /**
      * Zeros *only the angle* of the robot's field-relative control system.
-     * This this has *no effect* on odometry.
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -224,26 +224,38 @@ public void resetPosition() {
 
     /**
      * Zeros *only the angle* of the robot's field-relative control system.
-     * This this has *no effect* on odometry.
+     * This also resets localization to match the rotated field.
      */
     public void resetFieldAngle() {
```
e3l, line 229: the more i think abt it, the more it seems to me that this method is really just resetRotation, a piece of resetPosition. I think we should reset both field-oriented-drive and poseEstimator's heading in both cases, and the difference is in whether we set the translation? idk this is hurting my brain a lil, what do u/ya'll think

PR #41, opened 2023-02-06
Auto risplace

rishay-jain:


src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private enum Stage{
+        Engaged,
+        OnAxis,
+        Normal2Node, 
+        Ready,
+        Deposited
+    }
+
+    private Stage stage;
```
ky28059, line 43: Instead of another multi-stage command, consider chaining several calls with `.andThen()` or extending `SequentialCommandGroup` and calling `addCommands()`? `FollowPathCommand` already waits until it is at the position and angle so there's no reason to add a third layer of checking. I'm not sure what `OnAxis` means, but `FollowPathCommand` also handles waypoints so your segmented path can be better expressed as a single `FollowPathCommand` that goes from the current position to the Y-axis Pose2d using a waypoint, or if the angle matters a chain of two `FollowPathCommand`s using the above command chaining methods. The `Normal2Node` stage doesn't seem necessary either, and `Ready` can be another command in the sequence. Maybe you can look at how `BaseAutonSequence` is implemented.

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+    */ 
+    private final double BLUE_X_SET = 86.0; 
+    private final double BLUE_Y_SET = 30.0;
+    private final double BLUE_R_SET = -90.0; // rotation in degrees
+    
+    private final double RED_X_SET = 564.0; 
+    private final double RED_Y_SET = 30.0;
+    private final double RED_R_SET = 90.0; // rotation in degrees
+
+    Pose2d blueSetPoint = new Pose2d(BLUE_X_SET, BLUE_Y_SET, Rotation2d.fromDegrees(BLUE_R_SET));
```
ky28059, line 32: These can eventually be `PlacePosition`s like what caroline is implementing on the `crautonomous` branch? They also shouldn't be hard-coded (pass them in as params instead).

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.commands.swerve.FollowPathCommand;
+import frc.robot.subsystems.drivetrain.BaseDrivetrain;
+import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
+
+public class DepositCommand extends CommandBase {
+    private final BaseSwerveSubsystem driveSubsystem; // changed to BASE SWERVE SUBSYS to allow for getRobotPosition() 
+    private final AHRS ahrs; 
+
+    private final boolean red;
```
ky28059, line 16: I think this doesn't matter if you pass in positions to the constructor. If you're on the red team, you just pass in a red position.

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.commands.swerve.FollowPathCommand;
+import frc.robot.subsystems.drivetrain.BaseDrivetrain;
+import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
+
+public class DepositCommand extends CommandBase {
+    private final BaseSwerveSubsystem driveSubsystem; // changed to BASE SWERVE SUBSYS to allow for getRobotPosition() 
+    private final AHRS ahrs;
```
ky28059, line 14: Not needed?

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+        
+    }
+
+    // Called every time the scheduler runs while the command is scheduled.
+    @Override
+    public void execute() {
+
+        if(stage == Stage.Engaged){
+            if(red){
+                FollowPathCommand trajectory = new FollowPathCommand(
```
ky28059, line 68: How are you executing these commands if this isn't a `SequentialCommandGroup`?

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+                    driveSubsystem, 
+                    driveSubsystem.getRobotPosition(), 
+                    null,
```
ky28059, line 71: Use `List.of()` instead. Not sure if passing null causes a null pointer exception.

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+                    redSetPoint,
+                    true);
```
ky28059, line 73: `reverse` means that we drive backwards to the setpoint; does that apply here?

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+            
+            // insert run trajectory 
+
+        }
+
+        if(stage == Stage.OnAxis){ 
+
+            // driver input turned into pose2d along Y axis that robot goes to to deposit
+
+            // create next trajectory to that spot
```
ky28059, line 92: See above; this call can be chained in the constructor.

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        if(stage == Stage.Normal2Node){
+
+            // correct angle to multiple of 90º
```
ky28059, line 97: See above; `FollowPathCommand` does this automatically.

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+
+            // final vision-based alignment maybe?
+        }
+
+        if(stage == Stage.Ready){
+            // insert appropriate command based on selected deposit spot
```
ky28059, line 103: This can be done in the constructor with a `SequentialCommandGroup`.

src/main/java/frc/robot/commands/DepositCommand.java
```diff
@@ -0,0 +1,125 @@
...
+            // deposit
+        }
+
+        if(stage == Stage.Deposited){
+            // reset variables and robot and hand control back to driver
```
ky28059, line 108: This should be done in `end()`, if it needs to be done at all.

ky28059 (changes requested):

ky28059:
Superseded by #53.

PR #46, opened 2023-02-11
Fix-Tilt Hall-Effects & TeleOp Automations

e3l:


src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
```diff
@@ -142,10 +173,31 @@ public TiltedElevatorSubsystem() {
...
     }
 
+    
     @Override
     public void periodic() {
-        // if (zeroLimitSwitch.get()) extensionEncoder.setPosition(0); 
+        // Update elevator encoder using absolute magnet position
+        if (!hallEffectSensor.get()) {
+            // Use direction of elevator motion to set state of hall effect sensor
+            if (extensionEncoder.getPosition() > targetExtension) { // elevator moving down
```
e3l, line 182: this would check against the target extension of the previous cycle, which might not be problematic but doesn't seem quite right. I don't think targetExtension should be a field at all, I think it should be scoped to this method

e3l: also, i think this would trigger every periodic cycle that the hall effect sensor is on

src/main/java/frc/robot/motorcontrol/HallEffectSensor.java
```diff
@@ -0,0 +1,58 @@
...
+
+import edu.wpi.first.wpilibj.DigitalInput;
+
+public class HallEffectSensor {
+    private DigitalInput sensor;
+    private HallEffectMagnet[] magnets;
+
+    // Different values means between two magnets, same values means at a magnet.
+    private int lowerPos = -1; // -1 min, (array length - 1) max
+    private int upperPos = 0; // 0 min, array length max
```
e3l, line 11: im confused by what these represent

src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
pinkbluesky: i think this is outdated now

pinkbluesky:
rebased

src/main/java/frc/robot/commands/mover/TiltedElevatorCommand.java
```diff
@@ -1,16 +1,25 @@
...
+
     public TiltedElevatorCommand(TiltedElevatorSubsystem tiltedElevatorSubsystem, ElevatorState level) {
-        super(() -> {
-            System.out.println("Mover to " + level);
-            tiltedElevatorSubsystem.setState(level);
-            System.out.println("Mover at " + level);
-        }, tiltedElevatorSubsystem);
+        this.tiltedElevatorSubsystem = tiltedElevatorSubsystem;
+        System.out.println("Mover to " + level);
+        tiltedElevatorSubsystem.setState(level);
```
e3l, line 18: this should probably not happen in the constructor.. see `initialize()`?

src/main/java/frc/robot/motorcontrol/HallEffectMagnet.java
```diff
@@ -0,0 +1,27 @@
...
+    public HallEffectMagnet(double extendDistanceMeters, ElevatorState elevatorState) {
+        this.extendDistanceMeters = extendDistanceMeters;
+        this.elevatorState = elevatorState;
+    }
+    
+    public double getExtendDistanceMeters() { return extendDistanceMeters; }
+    
+    public double getExtendDistanceInches() { return Units.metersToInches(extendDistanceMeters); }
+
+    public boolean hasElevatorState() { return elevatorState != null; }
```
e3l, line 24: are we using this functionality?

src/main/java/frc/robot/subsystems/Superstructure.java
```diff
@@ -16,5 +17,6 @@ public Superstructure(RollerSubsystem rollerSubsystem, TiltedElevatorSubsystem t
     @Override
     public void periodic() {
         tiltedElevatorSubsystem.pieceGrabbed = rollerSubsystem.getPiece() != HeldPiece.EMPTY;
+        rollerSubsystem.allowOpen = tiltedElevatorSubsystem.getHeight() >= Units.inchesToMeters(20);
```
e3l, line 20: arbitrary, can probably add to constants?

src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
```diff
@@ -133,64 +144,101 @@ public TiltedElevatorSubsystem() {
...
 
+        // When magnet is detected, reset encoder
+        if (leftHallSensor != null) {
+            HallEffectMagnet leftHallSensorPos = leftHallSensor.getHallEffectState(extensionEncoder.getPosition());
+            // System.out.println(leftHallSensorPos);
+            // System.out.println(lastHallPos + " " + leftHallSensorPos);
+            if (leftHallSensorPos != null && lastHallPos == null){
+                REVLibError error = extensionEncoder.setPosition(Units.inchesToMeters(62.5));//leftHallSensorPos.getExtendDistanceMeters());
+                DriverStation.reportError(error.toString(), false);
+                DriverStation.reportError("bruh", false);
```
e3l, line 192: should probably remove this silly logging lol

src/main/java/frc/robot/commands/mover/TiltedElevatorCommand.java
penguin212: Fixed

src/main/java/frc/robot/motorcontrol/HallEffectMagnet.java
penguin212: Not currently, this is the logic for multiple magnets, which isn't needed at the moment. We can safely delete it if we are certain we won't be adding more magnets.

ky28059: If not using multiple magnets, you can probably delete the whole file and just use a `DigitalInput`?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -178,7 +179,11 @@ private void configureButtonBindings() {
             rollerSubsystem.setRollPower(rollPower);
         }, rollerSubsystem));
 
-        mechAButton.onTrue(new InstantCommand(rollerSubsystem::openMotor, rollerSubsystem));
+        if(driveSubsystem instanceof SwerveSubsystem){
```
ky28059, line 182: Does this need to be specifically a `SwerveSubsystem` and not just a `BaseSwerveSubsystem`? `SwerveSubsystem` doesn't define any additional methods, and you can just put this button binding in the `BaseSwerveSubsystem` if statement earlier in the method otherwise.

src/main/java/frc/robot/subsystems/Superstructure.java
penguin212: thumb up

src/main/java/frc/robot/commands/dropping/DropSequence.java
```diff
@@ -0,0 +1,57 @@
...
+         * 5. Moves backwards reverseDist
+         * 6. Waits waitTime3
+         * 7. Moves to GROUND
+         */
+
+        Pose2d currentPos = baseSwerveSubsystem.getRobotPosition();
+
+        Pose2d targetPos = currentPos.plus(new Transform2d(new Translation2d(-reverseDist * currentPos.getRotation().getCos(),
+                                                                             -reverseDist * currentPos.getRotation().getSin()),
+                                                                             new Rotation2d()));
```
ky28059, line 42: Passing `new Rotation2d()` to the `FollowPathCommand` means we always try to turn to 0 degrees even if the robot started at not 0 degrees. On the blue team, this means we always rotate to face away from the grid, while on the red team we always rotate to face it. We should probably just keep the initial angle?

src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
penguin212: done

src/main/java/frc/robot/motorcontrol/HallEffectMagnet.java
```diff
@@ -0,0 +1,27 @@
+package frc.robot.motorcontrol;
```
ky28059, line 1: These guys are still in `motorcontrol` when the folder was renamed to `util` a while back. Maybe move it there?

ky28059: idt it makes sense for magnets to be in a motor controller folder

src/main/java/frc/robot/motorcontrol/HallEffectSensor.java
```diff
@@ -0,0 +1,101 @@
...
+
+import edu.wpi.first.networktables.GenericEntry;
+import edu.wpi.first.wpilibj.DigitalInput;
+import edu.wpi.first.wpilibj.shuffleboard.BuiltInWidgets;
+import edu.wpi.first.wpilibj.shuffleboard.ShuffleboardTab;
+
+public class HallEffectSensor {
+    private DigitalInput sensor;
+    private HallEffectMagnet[] magnets; // array of magnet locations from smallest extension to largest extension
+    private ArrayList<GenericEntry> shuffleboardEntries = null;
```
ky28059, line 13: Can be a regular array maybe?

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -84,6 +89,19 @@ public void openMotor() {
         openTimer.start();
     }
 
+    public InstantCommand getOuttakeCommand(double power, double duration){
+        return(new InstantCommand(() -> {
+            roll(power, duration);
+        }));
+    }
+
+    public void roll(double power, double duration){
```
ky28059, line 98: If the above command is the only place where this method is used, it can be inlined. Also, why is constructing the command (and timer / states) within the subsystem necessary?

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -104,14 +122,23 @@ public void setRollPower(double power) {
...
-        openMotor.set(opening ? 0.3 : 0);
+        openMotor.set(opening && allowOpen ? 0.5 : 0);
 
         // If the cooldown has passed, stop and reset the timer.
         if (openTimer.hasElapsed(OPEN_TIME_SECONDS + COOLDOWN_SECONDS)) {
             openTimer.stop();
             openTimer.reset();
         }
 
+        if(rolling){
```
ky28059, line 133: The `rolling` state seems unnecessary; why not check if the timer has started with `.get() > 0` like the open motor logic does?

src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
```diff
@@ -237,4 +289,16 @@ public void resetOffset() {
     public void setManualPower(double power) {
         this.manualPower = power;
     }
+
+    public double getHeight() { 
+        return this.extensionEncoder.getPosition();
+    }
+
+    public double getTarget() {
```
ky28059, line 297: `getTargetExtension`? otherwise could be confused with `getState`

src/main/java/frc/robot/commands/dropping/DropSequence.java
penguin212: It does keep the initial angle, it is adding a vector to the current position with the .plus function.

src/main/java/frc/robot/motorcontrol/HallEffectMagnet.java
penguin212: It depends if we plan to add more magnets.

src/main/java/frc/robot/commands/dropping/DropperChooserCommand.java
```diff
@@ -32,27 +32,27 @@ private static SequentialCommandGroup getSequence(
         return switch (state) {
             case CONE_HIGH -> new DropSequence(
                 swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem,
-                ElevatorState.CONE_HIGH_DROP, .2, 0, 0, .2, 0, 0
+                ElevatorState.CONE_HIGH_DROP, .2, 0, 0, .2, 0
```
e3l, line 35: why do we need the first .2sec wait here / do we even need the CONE_HIGH_DROP enum? what if we added a getDropPos() to the state enum instead, and by default it's just the same position?

src/main/java/frc/robot/commands/dropping/DropSequence.java
```diff
@@ -30,28 +31,19 @@ public class DropSequence extends SequentialCommandGroup {
...
             new WaitCommand(waitTime1),
             new InstantCommand(rollerSubsystem::openMotor, rollerSubsystem),
             rollerSubsystem.getOuttakeCommand(-outtakePower, outtakeDuration),
             new WaitCommand(waitTime2),
-            // FollowPathCommand.from(swerveSubsystem, currentPos, List.of(), targetPos),
-            new WaitCommand(waitTime3),
+            new InstantCommand(() -> swerveSubsystem.setDrivePowers(-.2, 0, 0, true)),
+            new WaitCommand(backTime),
+            new InstantCommand(() -> swerveSubsystem.setDrivePowers(0, 0, 0, true)),
             new TiltedElevatorCommand(tiltedElevatorSubsystem, ElevatorState.GROUND)
```
e3l, line 47: i think this command can finish before the elevator is completely down, so the drivers can start driving again

src/main/java/frc/robot/commands/dropping/DropperChooserCommand.java
penguin212: This would require completely changing the implementation of TiltedElevator so that instead of passing in ElevatorStates for setting extensions we would just pass in an extension value (basically rewriting all of TiltedElevator). The current code only needs the special drop positions for 2 things, and the other version would be worse imo. 

I set the .2 to 0 (I think it was .2 from when the tiltedElevatorCommand didn't end).

src/main/java/frc/robot/commands/dropping/DropSequence.java
penguin212: Changed to an InstantCommand

src/main/java/frc/robot/commands/dropping/DropperChooserCommand.java
e3l: i'm saying that each enum of ElevatorState could have a built-in property that is its drop offset, which can be 0 by default

src/main/java/frc/robot/commands/mover/TiltedElevatorCommand.java
```diff
@@ -1,16 +1,35 @@
...
+        tiltedElevatorSubsystem.setState(targetState);
+        timer.start();
+    }
+
+    public void end(boolean interrupted) {
+        System.out.println("Mover at " + targetState);
+    }
+    
+    public boolean isFinished() {
+        return tiltedElevatorSubsystem.atTarget() || timer.hasElapsed(1);
```
e3l, line 33: do we need to tune this killswitch amt? would it be safer to use `withTimeout` so auto can use a different timeout?

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -8,12 +8,17 @@
...
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
-
 import frc.robot.util.MotorUtil;
 
 import static frc.robot.Constants.RollerConstants.*;
 
+import java.sql.Time;
+
+import org.opencv.osgi.OpenCVInterface;
+import org.opencv.osgi.OpenCVNativeLoader;
```
e3l, line 20: remove

src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
```diff
@@ -31,18 +36,18 @@ public class TiltedElevatorSubsystem extends SubsystemBase {
...
     private static final double EXTENSION_ROTATIONS_TO_METERS = EXTENSION_GEAR_RATIO * EXTENSION_CIRCUMFERENCE * 2.0 * (15.0 / 13.4);
-    private static final float EXTENSION_LIMIT = 26.0f;
 
     private static final double extensionP = 2.3;
     private static final double extensionI = 0;
     private static final double extensionD = 0;
     private static final double extensionTolerance = 0.003;
     private static final double extensionRampRate = 0.4;
-    private double arbFeedforward = 0.03;
+    private double arbFeedforward = 0;
```
e3l, line 50: only 0 if not IS_R1 from constants

src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
```diff
@@ -73,10 +83,14 @@ public double getExtension(boolean hasPiece) {
         },
         CHUTE(Units.inchesToMeters(40)),
         SUBSTATION(Units.inchesToMeters(50)), // absolute height = 37.375 in
-        CUBE_MID(Units.inchesToMeters(33)), // absolute height = 14.25 in
-        CUBE_HIGH(Units.inchesToMeters(53)), // absolute height = 31.625 in
-        CONE_MID(Units.inchesToMeters(50)), // absolute height = 34 in
-        CONE_HIGH(Units.inchesToMeters(0)); // absolute height = 46 in
+        CUBE_MID(Units.inchesToMeters(Constants.IS_R1 ? 33 : 40)), // absolute height = 14.25 in
+        CUBE_HIGH(Units.inchesToMeters(Constants.IS_R1 ? 53 : 55)), // absolute height = 31.625 in
+        CONE_MID(Units.inchesToMeters(48)), // absolute height = 34 in
```
e3l, line 88: let's keep the 50 tuned for R1 if IS_R1. same for CONE_HIGH

e3l (commented):
lgtm, we can probably merge so others can develop on top of these features and continue debugging/tuning on a new branch / main

i agree with kevin's earlier comments but let's implement those changes on new branches

src/main/java/frc/robot/commands/mover/TiltedElevatorCommand.java
penguin212: I'm going to change this to 3 seconds so that it is never used under normal operation, so it shouldn't affect auton. This way it only uses it if the elevator is taking absurdly long and is actually stuck.

PR #47, opened 2023-02-12
Auton unit testing

ky28059:
Very rough implementation at the moment, as shuffleboard doesn't seem to play well with simulation mode at all. I currently just have all the shuffleboard stuff commented out, but a better solution can probably be derived in the future. 

`PhotonWrapper` also didn't seem to like being reconstructed every test, but I realized I could just pass `null` to the swerve subsystem as it wouldn't be running periodic. Perhaps a more clean solution can be found for this too.

Less major of a concern, but the AHRS / phoenix framework startup logs kinda clog up the console when running tests.

(also the `BlueBalanceAuton` path is commented out in the test because it doesn't compile; this is a known issue and will be fixed soon)


ky28059:
Superseded by #51.

PR #50, opened 2023-02-13
Poll for PID updates with `.readQueue()`

ky28059:
SparkMax config IO operations like `pidController.setP()` should not be called every periodic loop, and constant reading and setting config parameters causes loop overruns. With NT3, the "correct" way to set PID gains only when shuffleboard values change would be to use callbacks, but NT4's pub-sub equivalent is `.readQueue()`.

> This uses the Set Parameter API and should be used infrequently.

Will need to be tested for performance before merge.

### DIFF
--- a/src/main/java/frc/robot/util/ShuffleboardUtil.java
+++ b/src/main/java/frc/robot/util/ShuffleboardUtil.java
@@ -3,6 +3,7 @@ package frc.robot.util;
 import java.util.Optional;
 import java.util.function.Consumer;
 
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX;
 import com.revrobotics.SparkMaxPIDController;
 
 import edu.wpi.first.networktables.GenericEntry;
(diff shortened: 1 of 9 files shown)


src/main/java/frc/robot/util/ShuffleboardUtil.java
```diff
@@ -0,0 +1,160 @@
...
+import edu.wpi.first.networktables.NetworkTableValue;
+
+public class ShuffleboardUtil {
+    /**
+     * Polls a Shuffleboard entry for a `double` change, updating the provided SparkMax PID controller's P gain if it is found.
+     * @param entry The shuffleboard entry to poll.
+     * @param pidController The PID controller to update.
+     */
+    public static void pollShuffleboardP(GenericEntry entry, SparkMaxPIDController pidController) {
+        pollShuffleboardValue(entry, (value) -> pidController.setP(value.getDouble()));
```
e3l, line 19: perhaps we can pass in a lambda to run if the value is changed instead of manually defining all these methods?

PR #51, opened 2023-02-14
Finished auton positions, mirroring

CrolineCrois:
…wards


src/main/java/frc/robot/commands/MotorTestCommand.java
```diff
@@ -0,0 +1,80 @@
...
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.CommandScheduler;
+import edu.wpi.first.wpilibj2.command.InstantCommand;
+
+import frc.robot.subsystems.RollerSubsystem;
+import frc.robot.subsystems.TiltedElevatorSubsystem;
+import frc.robot.subsystems.TiltedElevatorSubsystem.ElevatorState;
+import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
+
+public class MotorTestCommand extends InstantCommand {
```
e3l, line 12: maybe this shouldn't be an InstantCommand? i think currently this will lock up the CommandScheduler bc there are delays/sleeps in the command itself.

I think you could fix this by either using sequentialcommands or writing your own command that checks the time since starting and decides what to do based on that every time its `update()`ed.

src/main/java/frc/robot/commands/MotorTestCommand.java
```diff
@@ -0,0 +1,80 @@
...
+
+            // Run roller in and out
+            for (double i = -1 ; i <= 1; i += .1){
+                rollerSubsystem.setRollPower(i);
+                Timer.delay(0.2);
+            }
+            Timer.delay(TEST_DELAY_SECS);
+
+            // Open motor
+            rollerSubsystem.openMotor();
```
e3l, line 53: later we will have to use the DropCommand on the hall effect branch instead. we can probably run to a whole bunch of positions, end on `CONE_MID`, and test the dropping there bc it's most complex?

src/main/java/frc/robot/commands/grabber/RollerIntakeCommand.java
```diff
@@ -24,7 +24,7 @@ public void initialize() {
 
     @Override
     public void execute() {
-        rollerSubsystem.setRollPower(0.3);
+        rollerSubsystem.setRollPower(-0.3);
```
e3l, line 27: i think we should have negative power spit out and positive power suck in... is that not currently the case?

e3l (commented):
comments on pre-match checks

src/main/java/frc/robot/commands/sequences/BaseAutonSequence.java
```diff
@@ -38,24 +38,33 @@ public BaseAutonSequence(
...
      * @return The `SequentialCommandGroup` representing running the commands in order.
      */
-    protected Command goAndGrab(Pose2d initialPose, PiecePosition finalPose) {
-        return FollowPathCommand.from(swerveSubsystem, initialPose, List.of(), finalPose.getPose())
+    protected Command goAndGrab(Pose2d initialPose, List<Pose2d> waypoints, Pose2d finalPose) {
+        return FollowPathCommand.composedFrom(swerveSubsystem, initialPose, waypoints, finalPose)
             .alongWith(new TiltedElevatorCommand(tiltedElevatorSubsystem, ElevatorState.GROUND))
-            .alongWith(new RollerIntakeCommand(rollerSubsystem));
+            .andThen(new LockSwerveCommand(swerveSubsystem))
+            .andThen(new RollerIntakeCommand(rollerSubsystem).withTimeout(3));
```
e3l, line 49: it seems unlikely that the drivetrain will place us in a perfect positoin where running the intake will grab a piece. i'm almost certain we'll need to run the intake while driving into the piece to pick one up

src/main/java/frc/robot/positions/FieldPosition.java
```diff
@@ -0,0 +1,115 @@
+package frc.robot.positions;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.util.Units;
+
+import frc.robot.util.FieldUtil;
+
+public enum FieldPosition {
```
e3l, line 9: AutonPositions?

src/main/java/frc/robot/commands/swerve/LockSwerveCommand.java
```diff
@@ -0,0 +1,20 @@
...
+import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
+
+public class LockSwerveCommand extends InstantCommand {
+    public LockSwerveCommand(BaseSwerveSubsystem swerveSubsystem) {
+        super(() -> {
+            swerveSubsystem.setSwerveModuleStates(
+                new SwerveModuleState(0.0, new Rotation2d(Math.PI / 4.0)),
+                new SwerveModuleState(0.0, new Rotation2d(-Math.PI / 4.0)),
+                new SwerveModuleState(0.0, new Rotation2d(-Math.PI / 4.0)),
+                new SwerveModuleState(0.0, new Rotation2d(Math.PI / 4.0))
```
e3l, line 16: super nitpicky but this can probably be abstracted away so that we don't need a copy here and in (Abstract?)SwerveSubsystem

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -117,19 +117,19 @@ public RobotContainer() {
...
-            autonChooser.addOption("GRT path", new GRTAutonSequence(swerveSubsystem));
-
             autonChooser.addOption("Red top auton", new RedTopAuton(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
             autonChooser.addOption("Red balance auton", new RedBalanceAuton(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
             autonChooser.addOption("Red bottom auton", new RedBottomAuton(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
 
             autonChooser.addOption("Blue top auton", new BlueTopAuton(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
-            // autonChooser.addOption("Blue balance auton", new BlueBalanceAuton(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
+            autonChooser.addOption("Blue balance auton", new BlueBalanceAuton(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
             autonChooser.addOption("Blue bottom auton", new BlueBottomAuton(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
```
e3l, line 126: perhaps we can just add two Base(Top/Balance/Bottom)Auton classes, Red and Blue, for each path, instead of having explicit classes for each one? if the only difference is in passing a true/false we can probably do it here

e3l: or perhaps we can have two different choosers: auton path & red/blue

e3l (changes requested):

src/main/java/frc/robot/RobotContainer.java
ky28059: Two choosers won't work since the classes need the boolean when they're constructed, but I can def see the base autons no longer being abstract.

PR #55, opened 2023-02-19
Retune swerve modules, use limit switch on R1

ky28059:
Still need to retune the drifty module and might need to retune the binding module.

### DIFF
--- a/src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
@@ -155,7 +155,7 @@ public class TiltedElevatorSubsystem extends SubsystemBase {
 
     @Override
     public void periodic() {
-        if (zeroLimitSwitch.get()) extensionEncoder.setPosition(0); 
+        if (!zeroLimitSwitch.get()) extensionEncoder.setPosition(0); 
 
         // If we're in manual power mode, use percent out power supplied by driver joystick.
         if (IS_MANUAL) {
--- a/src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
@@ -6,7 +6,6 @@ import com.revrobotics.CANSparkMax.IdleMode;
 import com.revrobotics.CANSparkMax.SoftLimitDirection;
 import com.revrobotics.RelativeEncoder;
 import com.revrobotics.SparkMaxPIDController;
-import com.revrobotics.SparkMaxPIDController.AccelStrategy;
 import com.revrobotics.SparkMaxPIDController.ArbFFUnits;
 
 import edu.wpi.first.math.MathUtil;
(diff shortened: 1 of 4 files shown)


src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
```diff
@@ -157,8 +155,7 @@ public TiltedElevatorSubsystem() {
 
     @Override
     public void periodic() {
-        if (zeroLimitSwitch != null && zeroLimitSwitch.get())
-            extensionEncoder.setPosition(0); 
+        if (zeroLimitSwitch.get()) extensionEncoder.setPosition(0);
```
e3l, line 158: ! .get() !!!!

src/main/java/frc/robot/subsystems/drivetrain/MissileShellSwerveSubsystem.java
```diff
@@ -35,6 +41,32 @@ public MissileShellSwerveSubsystem() {
...
+
+            // The robot offset is the offset of the module assuming that it is aligned with the front of the robot.
+            // For the robot offset, subtract the position offset automatically applied by the module subclasses.
+            // currentAngle + offset + posOffset = 0 -> offset = -currentAngle - posOffset.
+            double robotOffset = MathUtil.angleModulus(-currentAngleRads - module.getPositionOffsetRads());
+            double flippedRobotOffset = MathUtil.angleModulus(-currentAngleRads - module.getPositionOffsetRads() + Math.PI);
+
+            // The pin offset is the offset of the module assuming that it is aligned with the aligning pin.
+            // For the pin offset, subtract pi/2 to account for the pin being 90 degrees off.
+            // currentAngle + offset + pi/2 = 0 -> offset = -currentAngle - pi/2.
```
e3l, line 58: this seems like a potential source of confusion--why don't we assume that the pin is the 'correct' 0, instead of 90deg off?

src/main/java/frc/robot/subsystems/drivetrain/SwerveModule.java
```diff
@@ -47,7 +47,7 @@ public class SwerveModule implements BaseSwerveModule {
     private static final double driveD = 0;
     private static final double driveFF = 0.176697057706;
 
-    private static final double steerP = 0.6;
+    private static final double steerP = 0.7;
```
e3l, line 50: did this lower the steady state error on each module?

ky28059: Honestly not sure, I think I only had one deploy with this before the back two modules were taken off. I can check tomorrow.

src/main/java/frc/robot/subsystems/drivetrain/MissileShellSwerveSubsystem.java
ky28059: Because our convention had been that the top left module should face the front of the robot at 0 degrees, but the pin makes it face 90. I thought that it'd just be easier to subtract pi/2 in desmos than to change all of our previous non-pin offsets.

PR #57, opened 2023-02-19
Prematch test command

ky28059:
@CrolineCrois


src/main/java/frc/robot/commands/MotorTestCommand.java
```diff
@@ -0,0 +1,80 @@
...
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.CommandScheduler;
+import edu.wpi.first.wpilibj2.command.InstantCommand;
+
+import frc.robot.subsystems.RollerSubsystem;
+import frc.robot.subsystems.TiltedElevatorSubsystem;
+import frc.robot.subsystems.TiltedElevatorSubsystem.ElevatorState;
+import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
+
+public class MotorTestCommand extends InstantCommand {
```
e3l, line 12: maybe this shouldn't be an InstantCommand? i think currently this will lock up the CommandScheduler bc there are delays/sleeps in the command itself.

I think you could fix this by either using sequentialcommands or writing your own command that checks the time since starting and decides what to do based on that every time its update()ed.

src/main/java/frc/robot/commands/MotorTestCommand.java
```diff
@@ -0,0 +1,80 @@
...
+
+            // Run roller in and out
+            for (double i = -1 ; i <= 1; i += .1){
+                rollerSubsystem.setRollPower(i);
+                Timer.delay(0.2);
+            }
+            Timer.delay(TEST_DELAY_SECS);
+
+            // Open motor
+            rollerSubsystem.openMotor();
```
e3l, line 53: later we will have to use the DropCommand on the hall effect branch instead. we can probably run to a whole bunch of positions, end on CONE_MID, and test the dropping there bc it's most complex?

src/main/java/frc/robot/commands/MotorTestCommand.java
```diff
@@ -0,0 +1,84 @@
...
+            new TiltedElevatorCommand(tiltedElevatorSubsystem, ElevatorState.SUBSTATION),
+            new WaitCommand(MECH_DELAY_SECS),
+
+            // TODO: run rollers in and out
+
+            // Open motor
+            new InstantCommand(() -> rollerSubsystem.openMotor()),
+            new WaitCommand(MECH_DELAY_SECS),
+
+            // Test swerve with S-curves
```
e3l, line 59: we do want to test swerve with simple directional movements / rotations so that it is easier to determine if everything is working correctly visually

e3l (approved):
lgtm!

PR #58, opened 2023-02-19
Flip cone command

ky28059:
@CrolineCrois

### DIFF
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -16,6 +16,7 @@ import edu.wpi.first.wpilibj2.command.RunCommand;
 import edu.wpi.first.wpilibj2.command.button.JoystickButton;
 
 import frc.robot.commands.BalancerCommand;
+import frc.robot.commands.FlipConeCommand;
 import frc.robot.commands.dropping.DropperChooserCommand;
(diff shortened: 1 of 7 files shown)


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -204,6 +205,8 @@ private void configureButtonBindings() {
         mechLBumper.onTrue(new InstantCommand(() -> {
             tiltedElevatorSubsystem.toggleState(ElevatorState.CONE_MID, ElevatorState.CONE_HIGH);
         }, tiltedElevatorSubsystem));
+
+        driveController.getBalancerButton().onTrue(new FlipConeCommand((BaseSwerveSubsystem) driveSubsystem, tiltedElevatorSubsystem));
```
e3l, line 209: seems like a weird button to use

src/main/java/frc/robot/subsystems/TiltedElevatorSubsystem.java
```diff
@@ -167,6 +168,11 @@ public void periodic() {
             return;
         }
 
+        if (flipCone){
```
e3l, line 171: i think instead of having this boolean we could add an elevator state like `SLOWLY_LOWER` and run custom logic if that's our target?

e3l (changes requested):

PR #60, opened 2023-02-19
Improve opening logic

ky28059:
This should work with the `allowOpen` logic. Conceptually, we should always be opening while the open timer is running and closing while the close timer is running. The moment `allowOpen` is set to false while we're opening, simply terminate the open timer as if it had finished naturally and start the closing timer.


ky28059:
Closed and reopened because rebase.

PR #61, opened 2023-02-20
Improve opening logic

ky28059:
See #60.

### DIFF
--- a/src/main/java/frc/robot/subsystems/RollerSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/RollerSubsystem.java
@@ -12,6 +12,7 @@ import edu.wpi.first.wpilibj.util.Color;
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
 
 import frc.robot.util.MotorUtil;
+import frc.robot.util.TrackingTimer;
 
 import static frc.robot.Constants.RollerConstants.*;
 
@@ -32,8 +33,8 @@ public class RollerSubsystem extends SubsystemBase {
     public boolean allowOpen = true;
     private boolean prevAllowedOpen = true;
 
-    private final Timer openTimer = new Timer();
-    private final Timer closeTimer = new Timer();
+    private final TrackingTimer openTimer = new TrackingTimer();
+    private final TrackingTimer closeTimer = new TrackingTimer();
 
     // ~110 for nothing, 130 for cone, 160 for cube
     private static final int CONE_PROXIMITY_THRESHOLD = 115;
@@ -113,7 +114,7 @@ public class RollerSubsystem extends SubsystemBase {
         prevAllowedOpen = allowOpen;
 
         // If we're opening and not allowed to anymore, or if we've finished opening, stop opening and start closing.
-        if ((openTimer.get() > 0 && stopOpening) || openTimer.hasElapsed(OPEN_TIME_SECONDS)) {
+        if ((openTimer.hasStarted() && stopOpening) || openTimer.hasElapsed(OPEN_TIME_SECONDS)) {
             openTimer.stop();
             openTimer.reset();
             closeTimer.start();
@@ -126,10 +127,9 @@ public class RollerSubsystem extends SubsystemBase {
         }
 
         // Otherwise, open if we're opening and close if we're closing.
-        openMotor.set(
-            openTimer.get() > 0 ? 0.5 :
-            closeTimer.get() > 0 ? -0.2 : 0
-        );
+        if (openTimer.hasStarted()) openMotor.set(0.5);
+        else if (closeTimer.hasStarted()) openMotor.set(-0.2);
+        else openMotor.set(0);
 
         // if wheels must intake, and the limit switch is not pressed, turn on motors
         if (limitSwitch.get()) {
--- a/src/main/java/frc/robot/subsystems/RollerSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/RollerSubsystem.java
@@ -35,6 +35,7 @@ public class RollerSubsystem extends SubsystemBase {
 
     private final TrackingTimer openTimer = new TrackingTimer();
     private final TrackingTimer closeTimer = new TrackingTimer();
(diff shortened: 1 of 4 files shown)


src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -129,34 +125,24 @@ public void setRollPower(double power) {
...
-
-        // If the cooldown has passed, stop and reset the timer.
-        if (openTimer.hasElapsed(OPEN_TIME_SECONDS + COOLDOWN_SECONDS)) {
-            openTimer.stop();
-            openTimer.reset();
-        }
+        // Otherwise, open if we're opening and close if we're closing.
+        openMotor.set(
+            openTimer.get() > 0 ? 0.5 :
+            closeTimer.get() > 0 ? -0.2 : 0
```
e3l, line 144: i feel that the use of the ternary here combined with the non-intuitive usage of the timer makes this a lil difficult to read

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -129,34 +125,24 @@ public void setRollPower(double power) {
...
         }
 
-        // If `OPEN_TIME` hasn't elapsed yet, run the motor.
-        // boolean opening = openTimer.get() > 0 && !openTimer.hasElapsed(OPEN_TIME_SECONDS);
-        // if (opening && allowOpen) {
-        //     openMotor.set(opening && allowOpen ? 0.5 : 0);
-        // }
-
-        // If the cooldown has passed, stop and reset the timer.
-        if (openTimer.hasElapsed(OPEN_TIME_SECONDS + COOLDOWN_SECONDS)) {
```
e3l, line 156: cooldown functionality seems to be removed completely

e3l (approved):

src/main/java/frc/robot/subsystems/RollerSubsystem.java
ky28059: What if we created a custom class that wrapped `Timer` to keep track of a `started` boolean and used `hasStarted()` here instead? Would that be easier to read?

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -90,12 +87,18 @@ public RollerSubsystem() {
     }
 
     /**
-     * Opens the roller by starting the open timer.
+     * Opens the roller by starting the open timer. Does nothing if we're not allowed
+     * to open, or if opening is still on cooldown.
      */
     public void openMotor() {
+        if (!allowOpen) return;
```
e3l, line 94: a later addition on main allowed for only a falling edge of allowOpen to trigger a close, so that the driver can still open the intake if they're deliberately doing so while underneath the allowed threshold

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -129,34 +132,24 @@ public void setRollPower(double power) {
...
-
-        // If the cooldown has passed, stop and reset the timer.
-        if (openTimer.hasElapsed(OPEN_TIME_SECONDS + COOLDOWN_SECONDS)) {
-            openTimer.stop();
-            openTimer.reset();
-        }
+        // Otherwise, open if we're opening and close if we're closing.
+        if (openTimer.hasStarted()) openMotor.set(0.5);
+        else if (closeTimer.hasStarted()) openMotor.set(-0.2);
+        else openMotor.set(0);
```
e3l, line 152: it might be tricky to integrate this with piece detection & powered clamping, if we use the latter solution. the idea would be to only use the motor to clamp actively (e.g. apply low voltage) if we're holding a cone, otherwise use 0 power.

e3l:
@ky28059 do you still think we should merge this? we might have a small window right now before the software needs to be changed for new intake. if so, please rebase?

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -111,35 +116,29 @@ public void periodic() {
...
-        boolean opening = openTimer.get() > 0 && !openTimer.hasElapsed(OPEN_TIME_SECONDS);
-        boolean closing = !opening || stopOpening;
-
-        if (closing) {
+        // If we're opening and not allowed to anymore, or if we've finished opening, stop opening and start closing.
+        if ((openTimer.hasStarted() && stopOpening) || openTimer.hasElapsed(OPEN_TIME_SECONDS)) {
+            openTimer.stop();
+            openTimer.reset();
             closeTimer.start();
+            cooldownTimer.start();
```
e3l, line 124: i do agree it makes sense to start the cooldown after we stop opening, but i think the old logic included opening time. can we adjust the cooldown constant to match?

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -90,12 +91,16 @@ public RollerSubsystem() {
...
+     * Opens the roller by starting the open timer. Does nothing if opening is still on cooldown.
      */
     public void openMotor() {
+        if (cooldownTimer.hasStarted() && !cooldownTimer.hasElapsed(COOLDOWN_SECONDS)) return;
+
         openTimer.start();
         closeTimer.stop();
         closeTimer.reset();
+        cooldownTimer.stop();
+        cooldownTimer.reset();
```
e3l, line 103: perhaps we can also abstract the .stop and .reset

src/main/java/frc/robot/subsystems/RollerSubsystem.java
ky28059: I thought the old logic didn't though; https://github.com/grt192/GRT2023/pull/61/files#diff-412251c91fef77914d9fc678673a89d2d22664ea9c1a90b42e59bb0bdcd98745L131-L132

e3l: oh okay you're right

PR #63, opened 2023-02-20
Charging station locking

ky28059:
Needs testing.


src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -220,6 +231,14 @@ private SwerveModulePosition[] getModuleStates() {
         };
     }
 
+    /**
+     * Sets whether the swerve should be locked parallel to the charging station.
+     * @param locked Whethe to lock the swerve parallel to the charging station.
```
ky28059, line 236: Typo will be fixed with the next logic change or before merge.

e3l (commented):
can probably add to a drive controller, also need to think about how this will work with the drivers moving around. perhaps can automatically ignore if drive inputs given, otherwise lock like this.

e3l (approved):

PR #64, opened 2023-02-20
Clean up hall effect sensor

ky28059:
Moves it from `motorcontrol` to `util`, uses a subclass for magnets, some other general cleanup.


ky28059:
Superseded by #83.

PR #67, opened 2023-02-21
low latency camera streams

e3l:


src/main/java/frc/robot/vision/SwitchableCamera.java
```diff
@@ -19,13 +21,26 @@ public SwitchableCamera(){
...
+        back.setBrightness(50);
+
         server = CameraServer.getServer();
     }
 
     /**
      * Switches the source connected to this camera (toggles between front and back).
      */
     public void switchCamera() {
+
```
ky28059, line 43: Newline!

ky28059 (approved):
No auto-flipping for the back camera?

e3l:
no automatic rotation for now because it appeared that the OpenCV pipeline increased latency significantly. 

we should look into making a custom widget that extends [CameraServerWidget](https://github.com/wpilibsuite/shuffleboard/blob/main/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java) or test OpenCV again with these updated camera settings.

PR #69, opened 2023-02-22
Add LED system for human player signalling

ky28059:
@pinkbluesky


e3l (commented):
add "blinking if we're holding a piece" logic?

e3l (approved):
it would be even cooler if it faded :))

also what about croline's rianbow commit, is there anything we want to keep from that

PR #70, opened 2023-02-22
Addition to DS logging

ky28059:
@rishay-jain

### DIFF
--- a/src/main/java/frc/robot/Logging/ErrorPacket.java
+++ b/src/main/java/frc/robot/Logging/ErrorPacket.java
@@ -1,0 +1,23 @@
+package frc.robot.Logging;
+
+import edu.wpi.first.math.geometry.Pose2d;
+
+public class ErrorPacket implements Message{
+    
+    private Pose2d position;
+    private String m;
+    private double velocity;
+
+    ErrorPacket(String m, Pose2d position, double velocity){
+        this.position = position;
+        this.velocity = velocity;
+        this.m = m;
+    }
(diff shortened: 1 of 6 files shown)


ky28059 (commented):
Is the idea to construct these packets in various subsystems depending on what is getting logged? How are these packets actually being logged? They seem to just be wrappers around strings at the moment.

rishay-jain:
they are wrappers around strings– whatever is printed out ends up in the driver station logs as well, so really its just a way of formatting printlns so they are more readable in the event list. it will also be able to automatically add in other pieces of information, such as position, heading, etc., so we get more context when reviewing logs.

ky28059:
So is it the responsibility of each subsystem to call `System.out.println(message.display())` or will there be some sort of manager structure? If it's the former, maybe this would be better as static methods on a util class like `LoggingUtil.logStateChange(...)` instead of having to construct a wrapper object just to call a formatting method and discard it.

ky28059:
ie. is the current plan to log a message from a subsystem
```java
StateChangePacket message = new StateChangePacket("...", "...", ..., ...);
System.out.println(message.display());
```
?

e3l:
https://www.chiefdelphi.com/t/improbable-java-slow-down-in-combining-doubles-and-strings/350331

ky28059:
I'd probably benchmark that to see if we're really having the same problem as them, because that thread is 4 years old and seems to be a Java 11 issue.

PR #66, opened 2023-02-21
Display relative / lock mode on Shuffleboard, vision enable toggle

ky28059:
A bit rough implementation-wise at the moment.

PR #73, opened 2023-02-23
Add VERBOSE_SHUFFLEBOARD flag

rishay-jain:


src/main/java/frc/robot/subsystems/drivetrain/SwerveModule.java
```diff
@@ -52,6 +52,8 @@ public class SwerveModule implements BaseSwerveModule {
     private static final double steerD = 0;
     private static final double steerFF = 0;
 
+    private boolean VERBOSE_SHUFFLEBOARD;
```
ky28059, line 55: I don't think this should be a constructor argument? Could just set it as a flag in either here or `Constants.java` so that you don't need to pass it around everywhere.

src/main/java/frc/robot/subsystems/drivetrain/SwerveModule.java
```diff
@@ -109,35 +111,49 @@ public SwerveModule(int drivePort, int steerPort, double offsetRads) {
...
+                .getEntry();
+            angleErrorEntry = shuffleboardTab.add("Angle error (rads)", 0.0)
+                .withPosition(6, 1)
+                .withSize(5, 3)
+                // .withWidget(BuiltInWidgets.kGraph)
+                .getEntry();
+        }
+        else{
+            shuffleboardTab = null;
+            targetVelEntry = null;
```
ky28059, line 150: Unless you modify the shuffleboard updates in `setDesiredState()` to have null checks as well, this will cause a runtime exception.

src/main/java/frc/robot/subsystems/drivetrain/SwerveModule.java
```diff
@@ -176,6 +192,7 @@ public void setDesiredState(SwerveModuleState state) {
         double currentVelocity = driveEncoder.getVelocity();
         double targetAngle = optimized.angle.getRadians() - offsetRads;
 
+
```
ky28059, line 195: newline

ky28059 (changes requested):

ky28059:
Closed in #66.

PR #76, opened 2023-02-24
Driver-Requested Features from practice & scrimmage

e3l:


src/main/java/frc/robot/vision/SwitchableCamera.java
```diff
@@ -17,10 +17,16 @@ public class SwitchableCamera {
...
+    private static final int FRONT_BRIGHTNESS = 80;
+
+    private static final int BACK_EXPOSURE = 40;
+    private static final int BACK_BRIGHTNESS = 80;
+
     private final VideoSink server;
     private final ComplexWidget widget;
 
-    private boolean frontActive = true;
+    private boolean topActive = true;
```
ky28059, line 29: But the cameras are still called `front` and `back`?

src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -235,6 +238,21 @@ public void setDrivePowers(double xPower) {
...
+    private ProfiledPIDController thetaController = new ProfiledPIDController(
+        0.5, 0, 0, 
+        new TrapezoidProfile.Constraints(
+            SwerveSubsystem.MAX_OMEGA / 2.,
+            SwerveSubsystem.MAX_ALPHA / 2.
+        )
+    );; // VERY MESSY!
+    public void setDrivePowersWithHeadingLock(double xPower, double yPower, Rotation2d heading, boolean relative) { // idk when to use Rotation2d and when to use a double
+        Rotation2d currentRotation = getRobotPosition().getRotation();
+        double error = MathUtil.angleModulus(heading.minus(currentRotation).getRadians());
```
ky28059, line 250: Doesn't `.minus()` already respect angle wrapping?

src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -235,6 +238,21 @@ public void setDrivePowers(double xPower) {
         setDrivePowers(xPower, 0.0, 0.0, true);
     }
 
+    private ProfiledPIDController thetaController = new ProfiledPIDController(
+        0.5, 0, 0, 
+        new TrapezoidProfile.Constraints(
+            SwerveSubsystem.MAX_OMEGA / 2.,
```
ky28059, line 244: Not super important for R2, but this can maybe use the constants already in the subsystem so that it also works for 2020?

src/main/java/frc/robot/controllers/BaseDriveController.java
```diff
@@ -27,6 +27,10 @@ public abstract class BaseDriveController {
      */
     public abstract boolean getSwerveRelative();
 
+    public boolean getHeadingLock() {
```
ky28059, line 30: Is this not an abstract method just so `TwistJoystickController` doesn't have to implement it?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -166,7 +173,15 @@ private void configureButtonBindings() {
                 double yPower = driveController.getLeftPower();
                 double angularPower = driveController.getRotatePower();
                 boolean relative = driveController.getSwerveRelative();
-                swerveSubsystem.setDrivePowers(xPower, yPower, angularPower, relative);
+
+                if (driveController.getHeadingLock()) {
+                    double currentHeading = swerveSubsystem.getRobotPosition().getRotation().getRadians();
+                    double lockHeading = (Math.abs(MathUtil.angleModulus(currentHeading)) > (Math.PI / 2)) ? Math.PI : 0;
```
ky28059, line 179: Somewhat confident that `currentHeading` is aways wrapped too, at least according to shuffleboard values when testing the field?

src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -235,6 +249,23 @@ public void setDrivePowers(double xPower) {
...
+     * 
+     * @param xPower The power [-1.0, 1.0] in the x (forward) direction.
+     * @param yPower The power [-1.0, 1.0] in the y (left) direction.
+     * @param heading The `Rotation2d` angle to lock the swerve to.
+     * @param relative Whether to use relative powers instead of field-oriented control.
+     */
+    public void setDrivePowersWithHeadingLock(double xPower, double yPower, Rotation2d heading, boolean relative) {
+        Rotation2d currentRotation = getRobotPosition().getRotation();
+        double error = MathUtil.angleModulus(heading.minus(currentRotation).getRadians());
+        double turnSpeed = thetaController.calculate(error);
```
ky28059, line 264: Oh another thing, we seem to be using a `ProfiledPIDController` which is profiled in units of omega (radians / second) and feeding its output directly to turnpower (which should be percent output between `[-1.0, 1.0]`)? Should we scale the calculated omega by dividing it by `MAX_OMEGA` first?

e3l: this explains a lot .

yheah lol

e3l: sorry i've been stuck in the "swerve uses velocities!!" mindset

src/main/java/frc/robot/Robot.java
```diff
@@ -44,6 +47,11 @@ public void disabledPeriodic() {}
 
     @Override
     public void autonomousInit() {
+        // disable vision
+        if (robotContainer.driveSubsystem instanceof BaseSwerveSubsystem) {
+            ((SwerveSubsystem) robotContainer.driveSubsystem).VISION_ENABLE = false;
```
ky28059, line 52: Cast to `BaseSwervesubsystem` instead? This will also make the shuffleboard vision enable entry not technically representative of the actual state of vision enabling, but that's lower priority; can be fixed by just making a method `setVisionEnable()` that changes both the boolean and the value of the entry.

src/main/java/frc/robot/Robot.java
```diff
@@ -59,6 +67,11 @@ public void teleopInit() {
         // continue until interrupted by another command, remove
         // this line or comment it out.
         if (autonomousCommand != null) autonomousCommand.cancel();
+
+        // enable vision
+        if (robotContainer.driveSubsystem instanceof BaseSwerveSubsystem) {
+            ((SwerveSubsystem) robotContainer.driveSubsystem).VISION_ENABLE = true;
```
ky28059, line 73: See above.

ky28059 (approved):

src/main/java/frc/robot/Robot.java
e3l: @ky28059 are you down to implement this?

PR #77, opened 2023-02-24
Interrupt balancer when auton is up to lock on station

ky28059:
Somewhat ugly implementation right now, but unsure how to improve.


e3l (approved):
lgtm if tested to work

PR #79, opened 2023-02-25
Balancing tuning

ky28059:
@rishay-jain

### DIFF
--- a/src/main/java/frc/robot/commands/balancing/PIDSwitchBalancerCommand.java
+++ b/src/main/java/frc/robot/commands/balancing/PIDSwitchBalancerCommand.java
@@ -81,4 +81,11 @@ public class PIDSwitchBalancerCommand extends BaseBalancerCommand {
         return reachedStation && balanced;
         // return reachedStation && Math.abs(ahrs.getPitch()) <= 2.0 && stoptimer.hasElapsed(0.20);
     }
+
+    double getPower(){
+        return returnDrivePower;
+    }
+    double getPitch(){
+        return ahrs.getPitch();
+    }
 }
--- a/src/main/java/frc/robot/commands/balancing/TuningBalancerCommand.java
+++ b/src/main/java/frc/robot/commands/balancing/TuningBalancerCommand.java
@@ -1,0 +1,43 @@
+package frc.robot.commands.balancing;
+
+import java.util.ArrayList;
+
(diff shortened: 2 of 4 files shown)


e3l:
@rishay-jain do we need this still?

src/main/java/frc/robot/commands/balancing/PIDSwitchBalancerCommand.java
```diff
@@ -81,4 +81,11 @@ public boolean isFinished() {
         return reachedStation && balanced;
         // return reachedStation && Math.abs(ahrs.getPitch()) <= 2.0 && stoptimer.hasElapsed(0.20);
     }
+
+    double getPower(){
```
ky28059, line 85: You don't have to manually implement this method in each subclass; inheritance will take care of it as long as it's in `BaseBalancerCommand.java`.

src/main/java/frc/robot/commands/balancing/TuningBalancerCommand.java
```diff
@@ -0,0 +1,43 @@
...
+import frc.robot.subsystems.drivetrain.BaseDrivetrain;
+
+public class TuningBalancerCommand extends DefaultBalancerCommand {
+
+
+    ArrayList<Double> time = new ArrayList<Double>();
+    ArrayList<Double> pitch = new ArrayList<Double>();
+    ArrayList<Double> power = new ArrayList<Double>();
+
+    Timer timer;
```
ky28059, line 15: Never initialized? Won't this throw `NullPointerException` at the moment?

src/main/java/frc/robot/commands/balancing/TuningBalancerCommand.java
```diff
@@ -0,0 +1,43 @@
...
+    public void execute(){
+        super.execute();
+        pitch.add(super.getPitch());
+        power.add(super.getPower());
+        time.add(timer.get());
+    }
+
+    @Override
+    public boolean isFinished(){
+        if(super.isFinished()){
```
ky28059, line 33: Seems like you don't need this method at all? These prints should really go in `end()` because you want them to run when the command ends, which means `isFinished()` should just return `super.isFinished()` and you can let inheritance handle it.

ky28059 (commented):
Also agree with Ethan about whether we still need this command. @rishay-jain

PR #80, opened 2023-02-25
Vision alignment

ky28059:
@rishay-jain

### DIFF
--- a/src/main/java/frc/robot/commands/grabber/VisionAlignmentCommand.java
+++ b/src/main/java/frc/robot/commands/grabber/VisionAlignmentCommand.java
@@ -1,0 +1,67 @@
+package frc.robot.commands.grabber;
+
+import org.photonvision.PhotonCamera;
+import org.photonvision.targeting.PhotonPipelineResult;
+import org.photonvision.targeting.PhotonTrackedTarget;
+
+import edu.wpi.first.math.controller.PIDController;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+
+import frc.robot.subsystems.RollerSubsystem;
+import frc.robot.subsystems.RollerSubsystem.HeldPiece;
+import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
+import frc.robot.vision.PhotonWrapper;
+
+/**
+ * Intakes 1 game piece with the roller mech. This command runs the rollers at a set power
+ * until the limit switch is triggered.
+ */
+public class VisionAlignmentCommand extends CommandBase {
+    private final BaseSwerveSubsystem driveSubsystem;
+
+    private double angularPower;
+    private final PIDController turnPID;
+    PhotonCamera recognitionCamera;
+
+    PhotonPipelineResult result;
+    boolean hasTargets;
+    PhotonTrackedTarget target;
+
+    public VisionAlignmentCommand(BaseSwerveSubsystem driveSubsystem){
+        this.driveSubsystem = driveSubsystem;
+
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/commands/grabber/VisionAlignmentCommand.java
```diff
@@ -0,0 +1,67 @@
...
+
+    PhotonPipelineResult result;
+    boolean hasTargets;
+    PhotonTrackedTarget target;
+
+    public VisionAlignmentCommand(BaseSwerveSubsystem driveSubsystem){
+        this.driveSubsystem = driveSubsystem;
+
+        recognitionCamera = new PhotonCamera("Microsoft_LifeCam_HD-3000");
+        turnPID = new PIDController(0.2/35, 0.0, 0.0);
```
ky28059, line 34: I'd use a `ProfiledPIDController` for this; see `FollowPathCommand`.

src/main/java/frc/robot/commands/grabber/VisionAlignmentCommand.java
```diff
@@ -0,0 +1,67 @@
...
+ * until the limit switch is triggered.
+ */
+public class VisionAlignmentCommand extends CommandBase {
+    private final BaseSwerveSubsystem driveSubsystem;
+
+    private double angularPower;
+    private final PIDController turnPID;
+    PhotonCamera recognitionCamera;
+
+    PhotonPipelineResult result;
```
ky28059, line 26: Doesn't have to be a state?

src/main/java/frc/robot/commands/grabber/VisionAlignmentCommand.java
```diff
@@ -0,0 +1,67 @@
...
+    boolean hasTargets;
```
ky28059, line 27: Also doesn't have to be a state

src/main/java/frc/robot/commands/grabber/VisionAlignmentCommand.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void initialize() {
+    } 
+
+    @Override
+    public void execute() {
+        result = recognitionCamera.getLatestResult();
+        hasTargets = result.hasTargets();
+        target = result.getBestTarget();
```
ky28059, line 46: What does this object look like if `hasTargets` is false? Does the command end execution at all if that happens?

src/main/java/frc/robot/commands/grabber/VisionAlignmentCommand.java
ky28059: Maybe after #76 gets tuned / merged this can use the `setDrivePowersWithHeadingLock()` method and not have to implement its own PID as well.

src/main/java/frc/robot/commands/grabber/VisionAlignmentCommand.java
```diff
@@ -0,0 +1,67 @@
...
+import frc.robot.vision.PhotonWrapper;
+
+/**
+ * Intakes 1 game piece with the roller mech. This command runs the rollers at a set power
+ * until the limit switch is triggered.
+ */
+public class VisionAlignmentCommand extends CommandBase {
+    private final BaseSwerveSubsystem driveSubsystem;
+
+    private double angularPower;
```
ky28059, line 22: Also doesn't have to be a state; you only need to store things in fields if you need to reference them across multiple methods or you need the previous value when rerunning the same method.

PR #81, opened 2023-02-26
Don't hit game piece in 1-piece autons

ky28059:
@CrolineCrois


e3l (approved):
does this need to be tested?

PR #83, opened 2023-02-28
Utah regional

ky28059:
The constant and whether optimization / wrapping work with the relative encoder needs to be tested.


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -191,7 +191,10 @@ private void driveBindings() {
             mechAButton.onTrue(new DropperChooserCommand(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
 
             driveController.getFieldResetButton().onTrue(new InstantCommand(swerveSubsystem::resetFieldAngle, swerveSubsystem));
-            driveController.getChargingStationLockButton().onTrue(new InstantCommand(swerveSubsystem::toggleChargingStationLocked, swerveSubsystem));
+            driveController.getChargingStationLockButton()
+                .onTrue(new InstantCommand(swerveSubsystem::toggleChargingStationLocked, swerveSubsystem));
+            driveController.getSwerveRelativeFeedbackButton()
```
e3l, line 196: i don't think this should go on the drive controller, i think it should go on the switchboard

src/main/java/frc/robot/subsystems/drivetrain/SwerveModule.java
```diff
@@ -104,7 +106,11 @@ public SwerveModule(int drivePort, int steerPort, double offsetRads) {
 
             steerAbsoluteEncoder = sparkMax.getAnalog(SparkMaxAnalogSensor.Mode.kAbsolute);
             steerAbsoluteEncoder.setPositionConversionFactor(STEER_VOLTS_TO_RADIANS);
-    
+
+            steerRelativeEncoder = sparkMax.getEncoder();
+            steerRelativeEncoder.setPositionConversionFactor(STEER_ROTATIONS_TO_RADIANS);
+            steerRelativeEncoder.setPosition(steerAbsoluteEncoder.getPosition()); // Set initial position to absolute value
```
e3l, line 112: the position of the steer motor will not be limited to [0, 2pi]... will simply calling `steerPidController.setReference` be able to handle this?

ky28059: Not sure, will need to test. Ideally, the PID wrapping will be able to handle this? But if it doesn't ig we can copy over the old optimization-with-wraparound code and conditionally apply it based on the feedback sensor being used.

e3l: i think we will need this. can we put it in, especially bc we won't really have time to test and we have a known working solution? we can leave the wrapping stuff commented in case it works, perhaps

ky28059:
@e3l are we merging this pull request?

src/main/java/frc/robot/commands/auton/BalanceAndTaxiAutonSequence.java
```diff
@@ -0,0 +1,49 @@
...
+import frc.robot.commands.balancing.DefaultBalancerCommand;
+import frc.robot.commands.balancing.GoOverCommand;
+import frc.robot.positions.FieldPosition;
+import frc.robot.positions.PlacePosition;
+import frc.robot.positions.PlacePosition.PlaceState;
+import frc.robot.subsystems.RollerSubsystem;
+import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
+import frc.robot.subsystems.tiltedelevator.TiltedElevatorSubsystem;
+
+public class BalanceAndTaxiAutonSequence extends BaseAutonSequence {
```
penguin212, line 15: Probably name this `TaxiAndBalanceAutonSequence` since we are taxiing first?

src/main/java/frc/robot/subsystems/tiltedelevator/TiltedElevatorSubsystem.java
```diff
@@ -218,7 +218,7 @@ public void periodic() {
         // else {
             // Set PID reference
             extensionPidController.setReference(
-                MathUtil.clamp(targetExtension, 0, EXTENSION_LIMIT),
+                MathUtil.clamp(targetExtension, 0, EXTENSION_LIMIT + Units.inchesToMeters(.75)),
```
penguin212, line 221: Is this not redundant to the soft limit? Clamping here is fine but maybe have two class variables `topLimit` and `botLimit` so that we can assure they are always set to the same value (in case we change one and forget to change the other).

src/main/java/frc/robot/commands/balancing/DefaultBalancerCommand.java
```diff
@@ -16,18 +16,24 @@ public class DefaultBalancerCommand extends BaseBalancerCommand {
...
     private double oldPitch;
     private double currentPitch;
     private double deltaAngle;
 
+    /**
+     * Constructs a default balancer command from a drive subsystem and a boolean indicating
+     * whether to balance from the front or back.
+     * 
+     * @param driveSubsystem The drive subsystem.
+     * @param reversed Whether to balance driving forwards or backwards. Defaults to backwards; pass `true` to balance forwards.
```
ky28059, line 31: The javadoc seems to have been left in; after this PR we can look into reversed balancing and maybe a better way of abstracting it?

src/main/java/frc/robot/commands/balancing/DefaultBalancerCommand.java
```diff
@@ -65,6 +70,7 @@ public void execute() {
             }
         }
 
+        // TODO: better name
```
ky28059, line 73: Another vestigial comment from when reverse powers were deleted 😳

src/main/java/frc/robot/commands/balancing/PIDSwitchBalancerCommand.java
```diff
@@ -21,9 +21,6 @@ public class PIDSwitchBalancerCommand extends BaseBalancerCommand {
     private double deltaAngle;
 
     private boolean reachedStation;
-    private boolean passedCenter;
-    private boolean waiting;
```
ky28059, line 25: I deleted all these boolean states because they weren't used, is there a plan to use them in the future @rishay-jain or are they vestigial?

src/main/java/frc/robot/commands/pretest/MotorTestCommand.java
```diff
@@ -23,10 +24,9 @@ public MotorTestCommand(BaseSwerveSubsystem swerveSubsystem, TiltedElevatorSubsy
         addRequirements(swerveSubsystem, tiltedElevatorSubsystem, rollerSubsystem);
 
         addCommands(
-            // lock swerve
-            new InstantCommand(() -> {
-                swerveSubsystem.applyLock();
-            }, swerveSubsystem),
+            new PrintCommand("hello 1"),
```
ky28059, line 27: Not the debug prints from when the test sequence was scheduled but not run

src/main/java/frc/robot/commands/pretest/MotorTestCommand.java
```diff
@@ -60,7 +62,7 @@ public MotorTestCommand(BaseSwerveSubsystem swerveSubsystem, TiltedElevatorSubsy
 
             // Test dropper command from `CONE_MID`
             DropperChooserCommand.getSequence(swerveSubsystem, rollerSubsystem, tiltedElevatorSubsystem),
```
ky28059, line 64: We deleted swerve because we wanted to run it on the cart, but the dropper command still runs swerve to back up doesn't it? Is the small duration it runs the wheels for ok?

src/main/java/frc/robot/subsystems/tiltedelevator/TiltedElevatorSubsystem.java
```diff
@@ -166,11 +166,11 @@ public void periodic() {
             // System.out.println(leftHallSensorPos);
             // System.out.println(lastHallPos + " " + leftHallSensorPos);
             if (leftHallSensorPos != null && lastHallPos == null){
-                extensionEncoder.setPosition(Units.inchesToMeters(EXTENSION_LIMIT));//leftHallSensorPos.getExtendDistanceMeters());
```
ky28059, line 169: I know this is commented out but `EXTENSION_LIMIT` is already in meters. Converting it as if it was in inches again would lead to an incorrect number; is that why it was commented?
```java
extensionEncoder.setPosition(leftHallSensorPos.getExtendDistanceMeters());
```
seems correct here.

src/main/java/frc/robot/commands/pretest/MotorTestCommand.java
e3l: was that ever fixed, by the way?

ky28059: No, but we can probably continue debugging on main.

src/main/java/frc/robot/commands/balancing/PIDSwitchBalancerCommand.java
rishay-jain: deleting them should be fine for now, I'll test those methods and add them back if needed

src/main/java/frc/robot/commands/pretest/MotorTestCommand.java
e3l: that's a good point... idk 😓

PR #85, opened 2023-03-02
Use heading lock PID for regular angular input

ky28059:
This should help eliminate "turning while translating" behavior caused by misbehaving modules.

### DIFF
--- a/src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
@@ -51,6 +51,7 @@ public abstract class BaseSwerveSubsystem extends BaseDrivetrain {
 
     private boolean chargingStationLocked = false;
 
+    private Rotation2d targetHeading = new Rotation2d();
     private Rotation2d angleOffset = new Rotation2d(0);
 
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -221,13 +220,44 @@ public void applyLock() {
      * @param relative Whether to use relative powers instead of field-oriented control.
      */
     public void setDrivePowers(double xPower, double yPower, double angularPower, boolean relative) {
+        // Increment target heading from angular power
+        this.targetHeading = targetHeading.plus(new Rotation2d(MAX_OMEGA * angularPower));
```
e3l, line 224: this essentially is saying that our target heading when full turn power is applied should be `MAX_OMEGA` away. this is an issue because `MAX_OMEGA` is rad/s but each loop timing is like 20ms.

e3l (commented):
we ran the robot very briefly before leaving and it seemed like the rotation issue was much smaller. with the current heading lock feature, it might be negligible. let's see how practice matches go tomorrow, but a major takeaway is that swerve offsets being off really does screw up how the drive acts

PR #86, opened 2023-03-08
Clean up balancer commands

ky28059:
Removed unnecessary states, duplicated calls to `ahrs.getPitch()`, and vestigial angle PID code (if we do want to lock it to an angle, we should use `setDrivePowersWithHeadingLock()` and probably lock it to 0 or 180 instead of the initial angle).

### DIFF
--- a/src/main/java/frc/robot/commands/balancing/DefaultBalancerCommand.java
+++ b/src/main/java/frc/robot/commands/balancing/DefaultBalancerCommand.java
@@ -3,6 +3,7 @@ package frc.robot.commands.balancing;
 import edu.wpi.first.wpilibj.DataLogManager;
 import edu.wpi.first.wpilibj.Timer;
 import edu.wpi.first.math.controller.PIDController;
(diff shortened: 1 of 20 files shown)


src/main/java/frc/robot/commands/balancing/DefaultBalancerCommand.java
```diff
@@ -49,6 +53,12 @@ public void execute() {
                 reachedStation = true;
                 returnDrivePower = -0.17;
             }
+            else if(runawayTimer.hasElapsed(2)){
+                returnDrivePower = 0.0;
+                System.out.println("BALANCER RUNAWAY DETECTED");
+                end(false);
```
ky28059, line 59: Calling `end(false)` won't actually end the command. You should set whatever `isFinished()` returns to `true`, and then the command scheduler will automatically end the command and call the end method.

e3l (approved):

e3l:
recommended to merge despite possible issues with balancing due to important non-balancing changes that need to be merged

PR #87, opened 2023-03-08
Color sensor tuning

penguin212:


ky28059 (commented):
What you can do instead of essentially manually implementing this is what we did last year for rejection logic in internals and use Rev's [`ColorMatch`](https://codedocs.revrobotics.com/java/com/revrobotics/colormatch) class. Add the cone, cube, and empty colors to the matcher with `.addColorMatch()` and either use `.matchClosestColor()` if the empty color is good enough or `.matchColor()` with some tuned confidence threshold if it's not.

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -187,23 +203,31 @@ private HeldPiece getColorSensorPiece() {
...
+        double emptyDist = Math.pow(emptyRed - red, 2) + Math.pow(emptyGreen - green, 2) + Math.pow(emptyBlue - blue, 2);
+        double coneDist = Math.pow(coneRed - red, 2) + Math.pow(coneGreen - green, 2) + Math.pow(coneBlue - blue, 2);
+        double cubeDist = Math.pow(cubeRed - red, 2) + Math.pow(cubeGreen - green, 2) + Math.pow(cubeBlue - blue, 2);
+        double minDist = Math.min(emptyDist, Math.min(coneDist, cubeDist));
+
+        System.out.println("empty -  " + emptyDist + " cone - " + coneDist + " cube - " + cubeDist);
         // TODO: modify `heldPiece` based on color sensor data
-        if (red > cubeRedMin && red < cubeRedMax && blue > cubeBlueMin && blue < cubeBlueMax && green < cubeGreenMax && green > cubeGreenMin){
-            // System.out.println("cube");
+        if (minDist == cubeDist){
```
ky28059, line 217: Also, doesn't using `==` instead of some kind of threshold make this require extremely precise readings? We probably won't get this consistency, especially when lighting changes between rooms / comps / etc.

penguin212: re-implemented for less comparisons

e3l (commented):
has this been tuned to 100% reliable?

if so, let's integrate with the other estimation methods (or replace proximity w/ this)

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -145,15 +160,17 @@ public void periodic() {
...
         HeldPiece limitPiece = getLimitSwitchPiece();
         HeldPiece proximityPiece = getProximitySensorPiece();
+        HeldPiece colorPiece = getColorSensorPiece();
 
         // If either the limit switch or the proximity sensor have detected a piece, set
         // the piece to the detected piece, prioritizing the proximity sensor over the limit
         // switch.
-        if (proximityPiece != HeldPiece.EMPTY) {
-            heldPiece = proximityPiece;
+        if (colorPiece != HeldPiece.EMPTY) {
```
ky28059, line 168: Do we want to do a 3-layer fallback here where we primarily use the color piece, fall back on the proximity piece, and finally the limit piece?

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -145,15 +160,17 @@ public void periodic() {
...
+            heldPiece = colorPiece;
         } else {
             heldPiece = limitPiece;
         }
+        getColorSensorPiece();
```
ky28059, line 173: Redundant call

src/main/java/frc/robot/subsystems/RollerSubsystem.java
e3l: alternatively, we shoudl remove call getProximityPiece.

layer of redundancy could be good but i'm also not sure if that'll take double the (already-long) time to poll the color sensor

PR #88, opened 2023-03-09
Auto align

ky28059:
Also includes a reabstraction of `PlacePosition` to use two `FieldPosition`s instead of reimplementing `FieldPosition`'s pose logic and abstracting the initial piece placement of the preloaded piece into `BaseAutonSequence`'s constructor.


src/main/java/frc/robot/commands/auton/BaseAutonSequence.java
```diff
@@ -71,36 +79,39 @@ protected Command goAndGrab(Pose2d initialPose, List<Pose2d> waypoints, Pose2d f
      * @param intialPose The initial pose of the robot.
      * @param waypoints The waypoints to hit along the path.
      * @param raisePose The pose to raise the elevator at.
-     * @param finalState The destination pose and elevator state of the robot.
+     * @param finalPose The destination pose of the robot.
+     * @param elevatorState The destination elevator state of the robot.
      * @return The `SequentialCommandGroup` representing running the commands in order.
      */
-    protected SequentialCommandGroup goAndPlace(Pose2d initialPose, List<Pose2d> waypoints, Pose2d raisePose, PlaceState finalState) {
+    protected SequentialCommandGroup goAndPlace(Pose2d initialPose, List<Pose2d> waypoints, Pose2d raisePose, Pose2d finalPose, ElevatorState elevatorState) {
```
e3l, line 86: why not just use a PlacePosition?

e3l (commented):
lgtm

src/main/java/frc/robot/commands/swerve/GoToPointCommand.java
```diff
@@ -0,0 +1,56 @@
...
+        );
+
+        // Pass powers and target heading to swerve subsystem
+        swerveSubsystem.setDrivePowersWithHeadingLock(
+            xPower, yPower, targetPose.getRotation(), false
+        );
+    }
+
+    @Override
+    public boolean isFinished() {
```
e3l, line 50: it might be worth adding a (configurable in the constructor) time kill switch also. if the driver can seamlessly take over it might not be useful in auto-aligning, but if we use this in auto, we'll probably want one

src/main/java/frc/robot/commands/dropping/AutoAlignCommand.java
```diff
@@ -0,0 +1,237 @@
...
+     */
+    public static PlacePosition getClosestPlacePosition(Pose2d robotPose, double elevatorExtensionMeters, boolean isRed) {
+        return Collections.min(
+            Arrays.asList(PlacePosition.values()),
+            Comparator.comparing((PlacePosition pos) -> {
+                Translation2d targetTranslation = pos.alignPosition.getPose(isRed).getTranslation();
+                return robotPose.getTranslation().getDistance(targetTranslation);
+            }).thenComparing((PlacePosition pos) -> {
+                double targetExtensionMeters =  pos.elevatorState.getExtension(OffsetState.DROPPING, true);
+                return Math.abs(targetExtensionMeters - elevatorExtensionMeters);
```
e3l, line 178: can we keep the current target elevator state instead? if the elevator is travelling to another state, it doesn't make sense to me to just take the closest one instead of getting to existing target

e3l:
to test thoroughly before merging this PR:
- auto aligning
- all utah autons: balancing (no mobility), one piece + taxi

src/main/java/frc/robot/commands/dropping/AutoAlignCommand.java
```diff
@@ -0,0 +1,239 @@
...
+
+        // Align with closest place position
+        scheduleAlignCommandWith(getClosestPlacePosition(initialPose, currentElevatorState, isRed));
+    }
+
+    /**
+     * Aligns to the node immediately left of the currently selected node (from the perspective of the drivers).
+     * This is a no-op if there isn't yet a node selected, or if the selected node is the left-most node already (C3).
+     */
+    public void alignLeft() {
```
e3l, line 100: "left" and "right" are probably misleading now. flipping based on alliance color doesn't seem to be implemented yet, either

src/main/java/frc/robot/commands/dropping/AutoAlignCommand.java
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * Returns a provided HYBRID, MID, or HIGH place position, depending on the given elevator state.
+     * @param elevatorState The current elevator state of the robot.
+     * @param hybridPosition The HYBRID position to return.
+     * @param midPosition The MID position to return.
+     * @param highPosition The HIGH position to return.
+     * @return The provided HYBRID, MID, or HIGH place position, depending on the given elevator state.
+     */
+    private static PlacePosition elevatorStateToPlacePosition(
```
e3l, line 150: this doesn't seem to be the most clean solution, but the alternative might be a big change. i'm thinking a class containing two enums. this would allow us to make methods that return a new instance of the class with one of the two enums different -- essentially, decouple the two states we're trying to decouple

e3l (approved):
lgtm, if not urgent:
- have we tested cancelling?
- need to test autons

known issues:
- photonvision config not persist

src/main/java/frc/robot/commands/swerve/GoToPointCommand.java
```diff
@@ -0,0 +1,82 @@
...
+        // Calculate swerve module states from desired chassis speeds, desaturating
+        // them to ensure all velocities are under MAX_VEL after kinematics.
+        SwerveModuleState[] states = kinematics.toSwerveModuleStates(speeds);
+        SwerveDriveKinematics.desaturateWheelSpeeds(
+            states, speeds, 
+            swerveSubsystem.MAX_VEL, swerveSubsystem.MAX_VEL, swerveSubsystem.MAX_OMEGA
+        );
+
+        // Pass states to swerve subsystem
+        swerveSubsystem.setSwerveModuleStates(states);
```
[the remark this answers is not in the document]
ky28059: Hmm, my thinking was that that would require two more overloads to be written for `.setDrivePowers` and `.setDrivePowersWithHeadingLock` and maybe it would make more sense for those two methods to act as the driver interface only (`FollowPathCommand` already uses `.setModuleStates()`, and the fact that `.setDrivePowers` takes [-1.0, 1.0] powers instead of velocities makes them more suited for joystick input than PID output)

ky28059:
Cancelling with the button definitely works, but unsure if Ryan used the joystick to cancel today.

src/main/java/frc/robot/commands/auton/BalanceAndTaxiAutonSequence.java
```diff
@@ -1,23 +1,20 @@
...
 import frc.robot.commands.balancing.DefaultBalancerCommand;
 import frc.robot.commands.balancing.GoOverCommand;
-import frc.robot.positions.FieldPosition;
 import frc.robot.positions.PlacePosition;
-import frc.robot.positions.PlacePosition.PlaceState;
 import frc.robot.subsystems.RollerSubsystem;
 import frc.robot.subsystems.drivetrain.BaseSwerveSubsystem;
 import frc.robot.subsystems.tiltedelevator.TiltedElevatorSubsystem;
 
 public class BalanceAndTaxiAutonSequence extends BaseAutonSequence {
```
CrolineCrois, line 13: taxi and balance

e3l:
tested balance & bottom blue autons

src/main/java/frc/robot/commands/auton/BalanceAndTaxiAutonSequence.java
e3l: yes true can u do this on ur branch

PR #89, opened 2023-03-11
LED control

ky28059:
@penguin212


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -114,10 +114,10 @@ public RobotContainer() {
         switchableCamera = new SwitchableCamera(shuffleboardTab);
 
         // driveSubsystem = new MissileShellSwerveSubsystem();
-        driveSubsystem = new SwerveSubsystem(photonWrapper);
+        signalLEDSubsystem = new LEDSubsystem();
```
e3l, line 117: can we move this above the missile, probably in its own section above, so the drives are together

src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -78,7 +79,8 @@ public BaseSwerveSubsystem(
         BaseSwerveModule bottomRightModule,
         double maxVel, double maxAccel, double maxOmega, double maxAlpha,
         SwerveDriveKinematics kinematics,
-        PhotonWrapper photonWrapper
+        PhotonWrapper photonWrapper,
+        LEDSubsystem ledSubsystem
```
e3l, line 83: perhaps we can add a constructor that takes no LEDSubsystem, leave the field as null, and have logic that's resilient to that. to minimize our impact, for example on SwerveSubsystem2020, and make it possible to not have LEDs

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -11,12 +11,17 @@ public class LEDSubsystem extends SubsystemBase {
 
     private final Timer blinkTimer;
     private static final double BLINK_DURATION_SECONDS = 0.5;
+    private final double APRIL_BLINK_DURATION = .2;
     private static final Color BLINK_COLOR = new Color(0, 0, 0);
     private boolean blinking = false;
+    private boolean continuous = false;
```
e3l, line 17: is this used anywhere? i'm struggling to read this new feature, it took me a bit to understand what continuous even meant, and i'm not sure how users are expected to interface with the methods in this class and also LEDStrip.java. can you please try documenting what methods are meant to be for, along with maybe your logic?

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -38,20 +43,70 @@ public void periodic() {
...
+        } else {
+            Color currentColor;
+            if(!aprilTimer.hasElapsed(APRIL_BLINK_DURATION)){
+                currentColor = aprilColor;
+            } else {
+                currentColor = color;
+            }
+            if(blinking){
+                ledStrip.setSolidColor(BLINK_COLOR);
+                ledStrip.updateContinuousColor(currentColor);
```
e3l, line 58: i think a source of confusoin is what `updateContinuousColor` does, and documenting that better might help. Why is it called after setting a solid color? Why is it called twice if blinking?

penguin212: i added documentation, update increments the colors while set actually sets the color. if you want to update the colors but not set them you only call update, otherwise you would call both.

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
penguin212: this is only used during the manual control mode, it is a toggle on the right joystick button

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -272,10 +273,23 @@ private void configureMechBindings() {
         }, tiltedElevatorSubsystem));
 
         signalLEDSubsystem.setDefaultCommand(new RunCommand(() -> {
-            if (blSwitch.getAsBoolean()) {
-                signalLEDSubsystem.setColor(255, 100, 0);
+            double x = mechController.getRightX();
+            double y = mechController.getRightY();
+
+            if(mechController.getRightStickButtonPressed()){
```
e3l, line 279: desired functionality is stick up = yellow, stick down = purple. if the stick is pressed down, color wheel

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -38,20 +44,97 @@ public void periodic() {
...
+            ledStrip.updateContinuousColor(color);
+            ledStrip.setContinuousColor();
+        } else if(manual){
+            //if manual and not continuous fill entire strip with color
+            ledStrip.fillContinuousColor(color);
+            ledStrip.setContinuousColor();
+        } else {
+            setColorPulse();
+            //setTwoColor()
+        }
```
e3l, line 58: we should not set the LEDs in both the periodic and in the `setRGB`/`setHSV` methods.

e3l (changes requested):

src/main/java/frc/robot/subsystems/leds/LEDStrip.java
```diff
@@ -26,4 +36,51 @@ public void setSolidColor(Color color) {
...
+        for(int i = 0; i < inc; i++){
+            contLedBuffer.setLED(i, color);
+        }
+        ledTimer.reset();
+        ledTimer.start();
+        
+    }
+
+    //set cont buffer on to the leds
+    public void setContinuousColor(){
```
ky28059, line 56: Is there a practical reason why the `setData(buffer)` call for continuous is separate from the "update buffer" call, and why the continuous buffer is separate from the solid buffer? It seems like right now every call to a "update continuous buffer" method is immediately proceeded by a "set buffer" call, and I think you can use the same buffer for both continuous mode and solid color since one simply overrides the other.

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -38,20 +50,116 @@ public void periodic() {
...
         this.color = new Color(
-            (int) (red * BRIGHTNESS_SCALE_FACTOR),
-            (int) (green * BRIGHTNESS_SCALE_FACTOR),
-            (int) (blue * BRIGHTNESS_SCALE_FACTOR)
+            (int) (r * BRIGHTNESS_SCALE_FACTOR),
+            (int) (g * BRIGHTNESS_SCALE_FACTOR),
+            (int) (b * BRIGHTNESS_SCALE_FACTOR)
+        );
+        //sets the whole strip to the color
+        ledStrip.fillContinuousColor(this.color);
```
ky28059, line 84: What Ethan said; the public methods should only modify the subsystem's state, and the LEDs should be set based on the state in `periodic()`.

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -38,20 +50,116 @@ public void periodic() {
...
+    public void setColorFromXandY(double x, double y){
+        if(manual){
+            double angleRads = MathUtil.inputModulus(Math.atan2(y, x), 0, 2 * Math.PI);
+            this.color = Color.fromHSV(
+            (int) (angleRads / (2 * Math.PI) * 180),
+            (int) 255,
+            (int) (255 * BRIGHTNESS_SCALE_FACTOR)
+        );
+        } else {
+            if(x > INPUT_DEADZONE){
```
ky28059, line 158: Small thing but you can merge this nested if with the `else` into an `else if` to improve readability. Generally the more nested layers there are in a single method the harder it gets to trace code execution flow.

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -38,20 +50,116 @@ public void periodic() {
...
+
+    public void toggleLEDControlMode(){
+        continuous = !continuous;
+    }
+
+    public void setManual(boolean manual){
+        this.manual = manual;
+    }
+
+    public void tagDetected(){
```
ky28059, line 111: Is there any documentation on what `aprilTimer` and `aprilTimer2` control? From what I can gather `aprilTimer` is used to send a color pulse and `aprilTimer2` controls the length of a 2-color display(?), but when is color pulse used and when is the two color system used? Are they ever used together?

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -38,20 +50,116 @@ public void periodic() {
...
+        }
+        if(!aprilTimer2.hasElapsed(.05)){
+            color2 = APRIL_COLOR;
+        } else {
+            color2 =  color;
+        }
+        ledStrip.setTwoColors(color, color2);
+    }
+
+    public void setColorFromXandY(double x, double y){
```
ky28059, line 149: Small thing again but would probably be more clear as `setDriverColor()` or something similar.

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -1,22 +1,34 @@
...
+    private final double APRIL_BLINK_DURATION = .2;
     private static final Color BLINK_COLOR = new Color(0, 0, 0);
     private boolean blinking = false;
+    private boolean continuous = false; //toggle in manual mode
+    public boolean manual = false; //when driver is directly controlling leds
 
     private static final double BRIGHTNESS_SCALE_FACTOR = 0.25;
+    private static final double INPUT_DEADZONE = .1;
 
     private Color color = new Color(192, 8, 254);
```
ky28059, line 26: If you're storing the colors as constants (which I think is a good idea), you might as well initialize this to `CUBE_COLOR`.

src/main/java/frc/robot/subsystems/leds/LEDStrip.java
penguin212: yes, often you want to set the buffer without moving it up.

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
penguin212: april timer and april timer 2 both are counting the time since an april tag was last seen

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -11,13 +12,23 @@ public class LEDSubsystem extends SubsystemBase {
...
     private static final double BLINK_DURATION_SECONDS = 0.5;
+    private final double APRIL_BLINK_DURATION = .1;
     private static final Color BLINK_COLOR = new Color(0, 0, 0);
     private boolean blinking = false;
+    public boolean manual = false; //when driver is directly controlling leds
 
     private static final double BRIGHTNESS_SCALE_FACTOR = 0.25;
+    private static final double INPUT_DEADZONE = .1;
 
     private Color color = new Color(192, 8, 254);
```
e3l, line 23: what is this used for?

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -11,13 +12,23 @@ public class LEDSubsystem extends SubsystemBase {
...
+    private final Color APRIL_COLOR = new Color(255, 0, 0);
+    private final Color CUBE_COLOR = new Color(192, 8, 254);
+    private final Color CONE_COLOR = new Color(255, 100, 0);
+    private final Timer aprilTimer = new Timer();
+    private final Timer aprilTimer2 = new Timer();
+    private Color lastColor = color;
     public boolean pieceGrabbed = false;
+    private boolean aprilStarted = false;
```
e3l, line 31: can u pls seperate all these variables on top into categories that are related to each other

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -38,20 +49,80 @@ public void periodic() {
...
+            aprilTimer.start();
+        }
+        aprilTimer2.reset();
+        aprilTimer2.start();
+    }
+
+    public void setColorPulse(){
+        //current color is the color that will be added at the bottom of the strip buffer
+        if(!aprilTimer.hasElapsed(APRIL_BLINK_DURATION) && aprilTimer.hasElapsed(.001)){
+            ledStrip.updateContinuousColor(APRIL_COLOR);
```
e3l, line 82: perhaps a more descriptive name is pushColorToContinuousBuffer?

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -38,20 +49,80 @@ public void periodic() {
...
+        } else {
+            ledStrip.fillContinuousColorIgnoringOneColor(color, APRIL_COLOR);
+            ledStrip.updateContinuousColor(color);
+        }
+        if(blinking){
+            ledStrip.fillContinuousColorIgnoringOneColor(BLINK_COLOR, APRIL_COLOR);
+        }
+        ledStrip.setContinuousColor();
```
e3l, line 90: this is the same as line 53 in LEDStrip, and it feels weird to me when the same logic is called by different parts of the code. perhaps LEDSubsystem can just always call this at the end of periodic? so everything before that is just responsible for modifying the buffer, and then it's sent to the LEDs at the very end of periodic

src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
e3l: ah, what if this was like a manualOverrideColor, and when it's null, it's not used and we do the default logic? that allows us to minimize states

PR #90, opened 2023-03-13
April tag false positive fix

penguin212:


ky28059 (changes requested):
Like Ethan said previously, it might not be the best idea to reject vision data based on current odometry readings because odometry might get messed up much more than a meter while out of range of april tags which, if we reject tag measurements more than a meter from the current pose, would effectively turn off vision readings for the rest of the match.

You could use a [`MedianFilter`](https://docs.wpilib.org/en/stable/docs/software/advanced-controls/filters/median-filter.html) to filter vision outliers without relying on odometry to detect them, or it might be the case that by increasing the decision matrix cutoff on photonvision that false positives can be eliminated there without any action on the Java side.

(also, this code would make it so that we would not be able to test auto-align or other position-dependent features without running auton since at robot bootup our odometry position is initialized to (0, 0) and any vision measurement is going to be in excess of a meter from that)

src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -419,4 +436,24 @@ public void setVisionEnabled(boolean visionEnable) {
...
+    public double minTargetDistance(List<PhotonTrackedTarget> trackedTargets){
+        Transform3d targetTransform = trackedTargets.get(0).getAlternateCameraToTarget();
+        double distance = Math.sqrt(Math.pow(targetTransform.getX(), 2) + Math.pow(targetTransform.getY(), 2));
+        double minAprilTagDist = distance;
+        
+        for(int i = 1; i < trackedTargets.size(); i++){
+            targetTransform = trackedTargets.get(i).getAlternateCameraToTarget();
+            distance = Math.sqrt(Math.pow(targetTransform.getX(), 2) + Math.pow(targetTransform.getY(), 2));
+            if(minAprilTagDist < distance){
+                minAprilTagDist = distance;
```
ky28059, line 449: If you wanted to get fancy with this method you can use `Collections.min()` with a custom comparator that mapped tracked targets to their `distance`s instead of writing a custom loop. You're already operating on a `List<T>`, so you wouldn't even need to convert an array to a `Collection<T>` for the operation.

src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -62,6 +68,9 @@ public abstract class BaseSwerveSubsystem extends BaseDrivetrain {
     private static final boolean SHUFFLEBOARD_ENABLE = true;
     private volatile boolean VISION_ENABLE = true;
 
+    private static final double APRIL_TAG_MAX_DIST_TO_ODOMETRY = 1.0; //meters 
+    private static final double APRIL_TAG_CUTOFF_DIST = 6.0; //meters
```
ky28059, line 72: What does this constant control?

src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -419,4 +436,24 @@ public void setVisionEnabled(boolean visionEnable) {
...
         visionEnableEntry.setBoolean(VISION_ENABLE);
     }
+
+    public double minTargetDistance(List<PhotonTrackedTarget> trackedTargets){
+        Transform3d targetTransform = trackedTargets.get(0).getAlternateCameraToTarget();
+        double distance = Math.sqrt(Math.pow(targetTransform.getX(), 2) + Math.pow(targetTransform.getY(), 2));
+        double minAprilTagDist = distance;
+        
+        for(int i = 1; i < trackedTargets.size(); i++){
+            targetTransform = trackedTargets.get(i).getAlternateCameraToTarget();
```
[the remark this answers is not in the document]
penguin212: What does this mean?

src/main/java/frc/robot/subsystems/drivetrain/BaseSwerveSubsystem.java
```diff
@@ -173,10 +179,11 @@ public void periodic() {
             if(ledSubsystem != null){
                 ledSubsystem.tagDetected();
             }
-            poseEstimator.addVisionMeasurement(
+            if(poseInField(visionPose.estimatedPose.toPose2d())){
```
e3l, line 182: super nitpicky but it might make sense to do a if !poseInField return; instead, only because it'll be easier to add other filters later on.

e3l (approved):
@ky28059 pls test

ky28059 (approved):
👍

PR #91, opened 2023-03-15
Modular auton

ky28059:
@CrolineCrois 

The idea with this setup is that `RobotContainer` uses shuffleboard choosers to select initial position, `isRed`, etc., and every time a value changes the robot would reconstruct the autonomous command (and auto-align when the team color changes) accordingly. We seemingly can't set up listeners on shuffleboard choosers though, so something else needs to be figured out for that.

Some TODOs from the current setup:
- Auton sequence is constructed at the start of auton in `.getAutonomousCommand()` instead of preconstructed before auton begins
- ~~Test if reconstructing `AutoAlignCommand` and rebinding the buttons successfully removes the old bindings~~


e3l (commented):
i also envision us being able to run "bottom one piece auton" and "top one piece auton" by selecting isRed - startPos - One Piece

PR #92, opened 2023-03-16
Multiply target velocity by cos(theta error)

ky28059:
Tested with no huge noticeable differences to odometry, though I believe it turns less while moving.


e3l (approved):
lgtm, though I'm concerned that this could affect driver experience... let's let ryan try it tomorrow

ky28059:
I didn't notice any different behavior when driving it manually yesterday, but I'll get Ryan to test it today for sure. For what it's worth, it seems to be a common thing teams do https://github.com/Mechanical-Advantage/RobotCode2023/blob/main/src/main/java/org/littletonrobotics/frc2023/subsystems/drive/Module.java#L116

PR #95, opened 2023-03-16
Constrain `GoToPointCommand` velocities using rate limiters

ky28059:


e3l (commented):
seems weird to use different limiters for x & y, though for our (almost-solely y-direction) movements i guess it's not a massive issue

e3l:
we seem to be making progress on auto-align using other solutions, so will close this for now

PR #99, opened 2023-03-17
some pfft intake stuff

penguin212:


src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -146,29 +151,48 @@ public void periodic() {
...
+        openMotor.setVoltage(-2);
+    }
 
+    protected void closeRoller(){
+        openMotor.set(-0.2);
+    }
+
+    protected void neitherRoller(){
+        openMotor.set(0);
+    }
```
e3l, line 174: these methods could probably be renamed to be more clear. for example i'm not sure if 'roller' should even be in the name.. unless roller means the intake as a whole?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -248,58 +249,58 @@ private void configureDriveBindings() {
      * Configures button bindings for the roller / elevator mechs and controller.
      */
     private void configureMechBindings() {
-        mechAButton.onTrue(new DropperChooserCommand(driveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
+        //mechAButton.onTrue(new DropperChooserCommand(driveSubsystem, rollerSubsystem, tiltedElevatorSubsystem));
 
         rollerSubsystem.setDefaultCommand(new RunCommand(() -> {
```
e3l, line 254: maybe we can break roller & elevator bindings into seperate functions so we can just comment out `configureElevatorBindings()` in the future.

e3l (approved):

PR #107, opened 2023-03-25
Color sensor reboot

e3l:


src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -231,6 +234,12 @@ private HeldPiece getColorSensorPiece() {
         gEntry.setValue(green);
         bEntry.setValue(blue);
 
+        //if the color is 0 0 0, the sensor is broken and should be rebooted
+        //only do this every couple seconds
+        if(colorTimer.advanceIfElapsed(2) && red == 0 && green == 0 && blue == 0){
```
e3l, line 239: this timer doesn't seem to be started. also we should immediately reboot if 0,0,0 but not try again until after 2 seconds if it's still broken

e3l (commented):
request changes

src/main/java/frc/robot/subsystems/RollerSubsystem.java
```diff
@@ -231,6 +233,20 @@ private HeldPiece getColorSensorPiece() {
...
         bEntry.setValue(blue);
 
+        //if the color is 0 0 0, the sensor is broken and should be rebooted
+        //only do this every couple seconds
+        if(red == 0 && green == 0 && blue == 0){
+            if(!colorTimer.hasStarted()){
+                crolorSensor = new ColorSensorV3(I2C.Port.kMXP);
+                colorTimer.start();
+            } else if (colorTimer.advanceIfElapsed(2)){
+                crolorSensor = new ColorSensorV3(I2C.Port.kMXP);
```
e3l, line 243: don't we also need to reset the timer here, otherwise it'll just spam after the 2nd time

e3l (commented):
request changes

PR #108, opened 2023-03-26
Don't reconstruct color sensor if not connected

ky28059:
If we lose communication with the sensor over I<sup>2</sup>C, we don't need to reinitialize the connection to start receiving data again; initializing the device only writes config settings to the color sensor. AFAIK I<sup>2</sup>C is not bidirectional and so we don't need to "handshake" before the sensor starts sending data to the RIO; rather, the RIO sends read requests to the sensor (when we call `.getColor()` in this case).

image

I think all reconstructing the sensor does is send unnecessary I<sup>2</sup>C write requests and spam console with errors.


e3l:
do these write to registers on the color sensor itself? if so, it might be necessary to reconnect to the color sensor -- this data might be lost when power is dropped.

PR #109, opened 2023-03-26
Color sensor signaling

penguin212:
Color sensor shows when it is dead. 
Intended behavior: if the color sensor is disconnected, 10 out of every 30 leds will turn red in groups along the led strip


src/main/java/frc/robot/subsystems/leds/LEDSubsystem.java
```diff
@@ -61,7 +64,14 @@ public void periodic() {
             //setTwoColor()
         }
 
-        ledStrip.setBuffer();
+        //if the color sensor is off, add red in groups on top of the current buffer, otherwise push the current buffer
+        if(colorSensorOff){
+            ledStrip.fillGroupedWithBlanks(COLOR_SENSOR_OFF_COLOR, 10, 20);
+            ledStrip.setOverlay();
```
e3l, line 70: i think this code would work but it's not immensely clean. the separation between LEDSubsytem and LEDStrip is quite confusing to me, and LEDStrip making the user manage overlay and everything seems odd too.

e3l:
tested to work @ paly

PR #112, opened 2023-03-27
Cross-fade between red and the base color on color sensor failure

ky28059:
This PR isn't quite able to fade into the manual color very effectively because of the increment logic; perhaps a larger "opacity" overhaul is needed.


ky28059:
Superseded by #114? Do we change that base into this branch and merge both or just close this PR?

penguin212:
Well we know this branch works, we should test #114 before merging it. We could merge this one now though.

PR #115, opened 2023-03-28
Fix theoretical bug with repeated aligning to same pose

ky28059:
If `currentPose` is queried by the `WaitUntilCommand` before a first call of `GoToPointCommand.execute()` (which it did at Paly, as seen by the `NullPointerException`), the old `currentPose` from the last run of the command will be stored and `isHeadingAligned` will immediately return true.

Therefore, theoretically aligning to the same point twice will cause the elevator to lift immediately without this behavior.


e3l:
do we not reconstruct the GoToPoint command everytime autoalign is triggered?

ky28059:
For the "align to closest" logic yes, for the buttons on shuffleboard no.

pinkbluesky:
Included in #116.

